PR #11, opened 2017-04-18
Implement OEM support

### LINKED ISSUE #9: Provide mechanism for OEM info [$50]
Some OEMs and distributors have mentioned that they'd like a place in the OS to put their contact information, i.e. in the About plug. This sounds like a good idea to me since our policy is generally that redistributors assume the support responsibilities for their customers.

I could see this implemented simply as a GSettings key to provide a distributor name and contact URL.

Launchpad Details: [#LP1582846](https://bugs.launchpad.net/bugs/1582846) Cassidy James Blaede - 2016-05-17 18:33:16 +0000

danrabbit:
fixes #9

### DIFF
--- a/src/Plug.vala
+++ b/src/Plug.vala
@@ -26,6 +26,12 @@ public class About.Plug : Switchboard.Plug {
     private string memory;
     private string graphics;
     private string hdd;
+    private bool oem_enabled;
+    private string manufacturer_name;
+    private string manufacturer_icon_path;
+    private string manufacturer_support_url;
+    private string product_name;
+    private string product_version;
     private Gtk.Label based_off;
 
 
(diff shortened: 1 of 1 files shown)


cassidyjames:
As a note to other reviewers, showing the OEM info requires an `/etc/oem.conf` file like this:

```
[OEM]
Manufacturer=System76 Inc.
Product=Meerkat
Version=meer1
Logo=/etc/oem/logo.png
URL=https://support.system76.com
```

as well as an image existing at the location specified by the `Logo` line.

src/Plug.vala
```diff
@@ -220,6 +226,35 @@ public class About.Plug : Switchboard.Plug {
             critical (e.message);
             hdd = _("Unknown");
         }
+
+        try {
```
cassidyjames: Ah, we'd talked about pretty-casing these instead of all caps, for more consistency with i.e. .desktop and other keyfiles.

danrabbit: done :)

PR #13, opened 2017-04-18
Show GTK version

### LINKED ISSUE #7: Show Gtk version info
Since elementaryOS freya ships with a Gtk version different than that of Ubuntu 14.04, I think that Gtk version info should also be shown in the About plug.

Launchpad Details: [#LP1404716](https://bugs.launchpad.net/bugs/1404716) Harshit - 2014-12-21 19:47:20 +0000

donadigo:
Fixes #7.


danrabbit:
screenshot from 2017-04-18 13 48 07

PR #23, opened 2017-05-04
Split hardware out into its own file

danrabbit:
I'd like to break up this big class into a couple more manageable classes. After this is merged I'll do a software view also

### DIFF
--- a/src/Views/HardwareView.vala
+++ b/src/Views/HardwareView.vala
@@ -218,15 +218,14 @@ public class About.HardwareView : Gtk.Grid {
         //and we want the <Info> part, we split with ":" and get the 3rd part
         string[] parts = graphics.split(":");
         string result = graphics;
-        if (parts.length == 3)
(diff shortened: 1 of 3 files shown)


src/Views/HardwareView.vala
```diff
@@ -0,0 +1,254 @@
...
+
+    private string get_graphics_from_string(string graphics) {
+        //at this line we have the correct line of lspci
+        //as the line has now the form of "00:01.0 VGA compatible controller:Info"
+        //and we want the <Info> part, we split with ":" and get the 3rd part
+        string[] parts = graphics.split(":");
+        string result = graphics;
+        if (parts.length == 3)
+            result = parts[2];
+        else if (parts.length > 3) {
```
Philip-Scott: Since you're touching the code anyways, you should fix the cuddled braces (And add some to that if up there) :)

PR #28, opened 2017-06-02
Implement Preferences URL Specification

peteruithoven:
Fixes: #3
The Preference Opening Specification document didn't mention a name, so I picked `about`.


danrabbit:
Thanks for your branch!

PR #32, opened 2017-06-20
Specify an icon pixel size for the OS Logo Icon

### LINKED ISSUE #31: OS Logo icon is very tiny if the current Icon theme doesn't specify 128px icons
If the current Icon theme doesn't explicitly provide 128px icons, then the OS Logo will use the hard-coded next available size down. This can be rather tiny in the case of an icon theme that designs up to a maximum of, say, 64px, but allows for scaling of those (SVG) icons up to larger sizes.
screenshot from 2017-06-20 13-50-06

isantop:
Fixes #31 

This PR specifies an pixel_size for the OS Logo Icon, which ensures that it's displayed at the right size even if an icon theme without 128px icons is used. The icon size matches the size that is originally requested, so there shouldn't be any conflicts with the elementary theme.


danrabbit:
Looks good, thanks for your branch :)

PR #33, opened 2017-06-30
Improve memory size displayed.

vjr:
Right now the memory size displayed is off a bit. This is an attempt to better display the actual physical memory installed with a more acurate rounded number so that layman users don't wonder where they lost or gained some memory. If the new method fails it falls back to the previous method to get the memory size.


src/Views/HardwareView.vala
```diff
@@ -245,7 +245,27 @@ public class About.HardwareView : Gtk.Grid {
...
+        int mempos = dmesg_line.index_of("Memory");
+        int start = dmesg_line.index_of("/", mempos);
+        int end = dmesg_line.index_of("K", start);
+        string size = dmesg_line.substring (start+1, end - start -1);
+        double total = double.parse (size);
+        int64 rounded = Math.llround (total / 1024.0 / 1024.0);
+        if (rounded == 0) {
+            return GLib.format_size (get_mem_info_fallback ());
+        }
+        return rounded.to_string () + " GB";
```
danrabbit: This won't work for RTL languages. Try `_"%s GB".printf (rounded.to_string ())`

vjr:
Some "before" and "after" screenshots on three separate computers - note the memory size shown. Ignore the "display" info items as that was from another branch.

Before:
about
about_envybird
about_firebird

After:
about_daily
about_mem_envybird
about_mem_firebird

vjr:
So I got a comment on Reddit to the effect:

> ... patch is unnecessarily complex. Reading dmesg output is prone to breaking due to kernel updates, and as it's a ring buffer the boot messages will be lost if the uptime is high enough. The actual problem with the original implementation is GLib.format_size (get_mem_info ()); which uses 1000 as the prefix instead on 1024. I think it could be fixed simply by using GLib.format_size_full with FormatSizeFlags.IEC_UNITS which uses 1024 as the prefix.

I'll see if this method is available and simply display a rounded integer rather than a decimal and see how it pans out.

vjr:
I just realised the IEC format while it may calculate in "GiB" instead of "GB" it might also display "GiB" instead of "GB" which might perplex some casual users. Also reading off `dmesg` seems to be get to a closer and more accurate memory size than `/proc/meminfo`, the code complexity notwithstanding.

cassidyjames:
How does this handle having non-integer numbers of GB? I.e., If I had 4608 MB, would it display 4.5 GB?

artemanufrij:
here is the command for get memory `dmesg -P | grep "Memory:" | grep -Po "(?<=/)(\d*)"`

donadigo:
@vjr I agree with the comment that has been issued at Reddit. This seems unnecessarily complex. This is essentially hardcoding to some executable and depending on it's string output, which could change in the future. If using `GLib.fromat_size` is not acceptable here, you should write a custom function that will format the `uint64 ` value returned by the original method.

vjr:
@cassidyjames for non integer results, like in your example it will divide by 1024 and round off to nearest integer so it will just show "4 GB" and not around "4.5 GB". But dont computers come with full GB intervals of RAM sizes (even older machines)?

cassidyjames:
@vjr many do, but it's definitely not enforced or standardized. And it would be _super_ weird if I had 4.5 GB RAM and the OS told me it had 4 GB.

danrabbit:
Gonna close this since no movement since January and this branch does some rounding which may yield incorrect results

PR #35, opened 2017-07-05
Show primary monitor information.

vjr:
Added a line to show the primary monitor's information, similar to the memory and storage lines.

It searches for the name and resolution and displays either or both or "Unknown" if not available.

See the following screenshots taken from three different computers:

about_display
about_envybird
about_firebird


danrabbit:
I’m -1 on this change as there is already a dedicated place to manage displays.

vjr:
Fair enough, it was still worth submitting the PR :-) I'll close this one.

peteruithoven:
I disagree, with @danrabbit isn't this relevant information about the computer? There are apps to manage disk space (Disks) and view memory usage (System monitor), although they are not included by default in ElementaryOS. 
I'm not sure if this is an argument, but the resulution is also included in `neofetch`. 

Other interesting display information: 
- Physical size (amount of inches)
- DPI / PPI?
- touch enabled?

vjr:
My opinion is to include this in the About plug but with the basic information like I showed in my screenshots and not go overboard with the details. Maybe the additional details like physical size, DPI/PPI, touch enabled could be added to the Display plug if warranted.

peteruithoven:
I understand you don't want to show too much in this screen. 
Maybe more info could be shown on hover? I mean the "(rev ...)" info on gpu's also seems a bit much & quite technical to be included by default.

btkostner:
Going to point to #36 as it's somewhat related.

danrabbit:
Yeah, I'm gonna "won't fix" this one. There is already a place for this information in the same app ;)

PR #46, opened 2018-01-13
Use laptop icon if we're on a laptop

danrabbit:
Steal the laptop detect code from switchboard-plug-power

### DIFF
--- a/src/Views/HardwareView.vala
+++ b/src/Views/HardwareView.vala
@@ -29,12 +29,19 @@ public class About.HardwareView : Gtk.Grid {
     private string product_name;
     private string product_version;
 
+    private const string LAPTOP_DETECT_NAME = "laptop-detect";
+
     construct {
         fetch_hardware_info ();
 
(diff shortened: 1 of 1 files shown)


src/Views/HardwareView.vala
```diff
@@ -263,4 +270,26 @@ public class About.HardwareView : Gtk.Grid {
...
+        if (laptop_detect_executable == null) {
+            warning ("laptop-detect not found");
+            return false;
+        }
+
+        int exit_status;
+
+        try {
+            Process.spawn_command_line_sync (laptop_detect_executable, null, null, out exit_status);
+            if (exit_status == 0) {
```
Bat41: You could just reduce it to `return exit_status == 0`

danrabbit: Admittedly this is copy/pasted from the power plug, but I think the purpose of doing it this way is for that debug message

Conan-Kudo:
@danrabbit What is `laptop-detect`?

PR #48, opened 2018-02-10
Fix layout

markodp:
* make sure that images are the same distance from separator
* added left margin to software and right margin to hardware so that grids are centered evenly between window edge and separator when you downsize window 
* use ellipsis on hardware labels so that window doesn't auto-resize if text is too big
* oem product was aligned on right if there was no oem version provided

### DIFF
--- a/src/Plug.vala
+++ b/src/Plug.vala
@@ -258,8 +258,8 @@ public class About.Plug : Switchboard.Plug {
         description_grid.valign = Gtk.Align.CENTER;
         description_grid.vexpand = true;
         description_grid.column_spacing = 24;
-        description_grid.margin_left = 12;
-        description_grid.margin_right = 12;
+        description_grid.margin_start = 12;
+        description_grid.margin_end = 12;
         description_grid.add (software_grid);
         description_grid.add (new Gtk.Separator (Gtk.Orientation.VERTICAL));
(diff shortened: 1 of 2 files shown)


markodp:
For compariosn, this was before:
screenshot from 2018-02-10 23 57 28

And now:
screenshot from 2018-02-10 23 58 12

jeremypw:
With current master on Juno, I do not get the asymmetric layout anyway - it already looks much as in your second image.  Your PR made very little difference - I only spotted that the ellipsization was slightly different.  Are you sure this PR is needed?

jeremypw:
Another difference found is that the window is not horizontally resizable in master but it is in PR.

peteruithoven:
For comparison a screenshot from a (unchanged) Loki with customized oem config:
screenshot from 2018-02-13 19 28 31

markodp:
I did test this on juno, screenshots are from there. This alignment fix would only be visible in some circumstances where there is not much text on the right hardware side, like in my case. Maybe try reducing long labels by editing them with GTK Inspector and you will see the difference this makes.

markodp:
I just created one for testing. When i removed version, product aligned on right.

Something like this:
```
[OEM]
Product=Product
Manufacturer=Manufacturer
Website=example.com
```

PR #49, opened 2018-02-15
Refactor Plug.vala

donadigo:
A couple of things to make the code cleaner:

* explicit types to `var`
* Space before brackets
* Code simplifications
* Move some methods to the class scope
* Make non-instance methods static


jeremypw:
Shouldn't reset schemas be looking for prefix "io.elementary" as well?

src/Plug.vala
```diff
@@ -273,80 +273,74 @@ public class About.Plug : Switchboard.Plug {
...
-private string[] get_pantheon_schemas () {
-    string[] schemas = {};
-    string[] pantheon_schemas = {};
-    string[] prefixes = { "org.pantheon.desktop", "org.gnome.desktop" };
+        return result == 1;
+    }
 
-    var sss = SettingsSchemaSource.get_default ();
+    private void settings_restore_clicked () {
+        var should_display = confirm_restore_action ();
```
jeremypw: Is this variable necessary?  Only used once and does not increase clarity significantly imo.

src/Plug.vala
```diff
@@ -273,80 +273,74 @@ public class About.Plug : Switchboard.Plug {
...
-
-    reset_all_keys (settings);
-
-    var children = settings.list_children ();
-    foreach (var child in children) {
-        var child_settings = settings.get_child (child);
 
-        reset_all_keys (child_settings);
+    private static void reset_all_keys (GLib.Settings settings) {
+        var keys = settings.list_keys ();
```
jeremypw: Is this variable necessary?  Only used once and does not increase clarity significantly imo.

src/Plug.vala
```diff
@@ -273,80 +273,74 @@ public class About.Plug : Switchboard.Plug {
...
+        return result == 1;
+    }
 
-    var sss = SettingsSchemaSource.get_default ();
+    private void settings_restore_clicked () {
+        var should_display = confirm_restore_action ();
 
-    sss.list_schemas (true, out schemas, null);
+        if (should_display) {
+            var all_schemas = get_pantheon_schemas ();
```
jeremypw: Is this variable necessary? Only used once and does not increase clarity significantly imo.

PR #54, opened 2018-03-11
Set center justify on graphics label

### LINKED ISSUE #53: Wrong alignment
As you can see at image, text block with info about video card is aligned by the left, not center like the rest

danrabbit:
Fixes #53 ... I think. Someone with multiple graphics cards needs to test


peteruithoven:
Before: 
screenshot before

After:
screenshot after

So seems fixed :)

PR #56, opened 2018-04-28
Report an Issue dialog

danrabbit:
Replaces the single report an issue link with a fancy dialog. Since there's no longer a single link to report an issue like there was with Launchpad, we have to know what the user intends to report an issue with first.

screenshot from 2018-04-28 14 28 51 2x
screenshot from 2018-04-28 15 12 01 2x


danrabbit:
@donadigo since it’s a messagedialog subclass, it’s already non-deletable :)

arshubham:
@danrabbit I like this very much. Any reason why only the system apps are included. Why not also include made for elementary apps installed from appcenter. A user may not know the difference between them. Maybe they can be included under a new heading of user installed apps.

danrabbit:
@arshubham Maybe we can look at that in a future revision. This branch is already a pretty good size and I'd like to land it without increasing it's scope of features

danrabbit:
Woo thanks :)

Issue #63, opened 2018-07-13
Show what kind of storage is used


ryonakano:
This may not be always necessary, but if we can check I suppose it's nice:
e.g. SATA HDD, SATA SSD, or NVMe SSD

marcelogdeandrade:
I'm working on this. The first solution I found was reading this file:

$ cat /sys/block/sda/queue/rotational

as explained in [this](https://unix.stackexchange.com/questions/65595/how-to-know-if-a-disk-is-an-ssd-or-an-hdd) post.

The problem is, it only works if the mounted device's name is "sda" but I think I can get around it looking for the devices's name before searching this file.

marcelogdeandrade:
@danrabbit I solved this issue of finding out the disk's name executing "$ df /". But I don't know if it is possible or recommended to execute a command inside an elementary application.

Can I take this approach or should I try to solve it by getting this information from a system file?

PR #67, opened 2018-08-20
Adds storage type to storage info

marcelogdeandrade:
Adds feature requested on issue #63.

As explained [here](https://unix.stackexchange.com/questions/65595/how-to-know-if-a-disk-is-an-ssd-or-an-hdd), this feature reads the rotational file from the disk's directory. If this file contains a 0, it is a HDD, if it contains a 1, it is a SSD.

The problem is that sometimes the disk name changes. The large majority of systems have "sda" as the block device that mounts /. This implementation only works in this case. If the system's block device is not "sda", no storage type is shown.


marcelogdeandrade:
I did a new commit to add the feature it was missing, detecting the disk's name. Now it should work in all systems. The way I did it was running two commands with "spawn_command_line_sync" from vala.

The first one is "df /" to get the partition's name that mounts /. The second one is "lsblk -no pkname /dev/sda", in this case the partition is sda, but it receives the result from the first command. Both commands are standard for linux.

danrabbit:
Thanks for your branch! I think we usually want to use libraries for thing like this though. In this case, it looks like you can use [`Gdu.Device.drive_get_is_rotational`](https://valadoc.org/gdu/Gdu.Device.drive_get_is_rotational.html)

marcelogdeandrade:
Thanks for your feedback!
This package is not available by default on Juno's aptitude manager, should I install an older .deb file?

danrabbit:
Oh oops, you're right. I think the replacement here might be [`UDisks.Drive.rotation_rate`](https://valadoc.org/udisks2/UDisks.Drive.rotation_rate.html)

marcelogdeandrade:
Ok, thanks, I'll try to update the branch using this lib.

marcelogdeandrade:
I changed my implementation to use udisks2 package. It was my first time using Vala packages, so i don't know if my changes were correct. I updated the readme, meson.build and travis file with it. 

I only tested it with my SSD drive, so I recommend testing it with a HDD drive to make sure it is working well.

src/Views/HardwareView.vala
```diff
@@ -280,6 +282,23 @@ public class About.HardwareView : Gtk.Grid {
 
         return 0;
     }
+
+    private string get_storage_type () {
+        string storage = "";
+        try {
```
danrabbit: There isn't an Errorable method here, so we actually get a warning from the compiler that the catch clause is unreachable. So this can be removed

src/Views/HardwareView.vala
```diff
@@ -280,6 +282,23 @@ public class About.HardwareView : Gtk.Grid {
...
+            UDisks.DriveSkeleton drive = new UDisks.DriveSkeleton ();
```
danrabbit: elementary code style here is to use `var drive` instead of `UDisks.DriveSkeleton drive`

src/Views/HardwareView.vala
```diff
@@ -280,6 +282,23 @@ public class About.HardwareView : Gtk.Grid {
...
+            int rotational = drive.rotation_rate;
+            if (rotational == 0) {
+                storage = "(SSD)";
```
danrabbit: These strings should probably be translatable. So it should be `storage = _("(SSD)");`. Don't forget the HDD as well :)

marcelogdeandrade:
I did the requested changes, thanks for the help. 

The dependencies on readme were not on alphabetic order (libswitchboard-2.0-dev before libgranite-dev), should I change this? Nevertheless, I added libudisks2 dependency on order without changing the current order.

marcelogdeandrade:
The travis build failed, but I could not find out why, it looks like some connections could not be made, I don't know if it is a problem with my changes

src/Views/HardwareView.vala
```diff
@@ -280,6 +282,19 @@ public class About.HardwareView : Gtk.Grid {
 
         return 0;
     }
+
+    private string get_storage_type () {
+        string storage = "";
+        var drive = new UDisks.DriveSkeleton ();
```
donadigo: As the class is named "DriveSkeleton" and it's description says `Creates a skeleton object for the D-Bus interface org.freedesktop.UDisks2.Drive`, it's not an actual implementation nor something that actually gets the information about the drive that contains the system partition. It doesn't know magically what drive to check, it's just a dummy object.

In this case, it's coincidental that everyone that tested this branch had an SSD. It shows SSD, because by default the `rotation_rate` is 0. This doesn't work currently with HDD's.

At this point you should try to use the [UDisks.Client](https://valadoc.org/udisks2/UDisks.Client.html) class and then try to enumerate on the drives, see which is the "/" and check the rotation_rate of that.

marcelogdeandrade:
Which implementation is better? To enumerate the drives using Udisks2 lib and than proceed to find out which one is mounting /, or do my previous implementation and get the partition name from / with "df /" without using external libs?

I haven't tried with Udisks2.Client, but I think in the end it will be the same effort as the first one, but using and external lib.

marcelogdeandrade:
I've been trying to implement this solution with Udisks2.Client without success. All methods seem to require some previous information to list disks,blocks or partitions (MRaid, a Drive or a Block). Getting this objects is not a easy task aswell, I've tried to use some Proxy objects but they all require a DBusConnection.

At this point, I think solving this problem using a lib is not easier than doing it manually. Should I roll back my commits to the last version when I solved it without libs?

marcelogdeandrade:
@danrabbit @donadigo I'm sorry to bother you guys again, but it's been 7 days without a response about this PR. I know this isn't any priority on elementary, I would just like to know if I should continue working on this or if it is better to close it and try something else. 

Thanks for your help!

donadigo:
@marcelogdeandrade sorry, I forgot about your response. Tbh I feel like using an entire  library is overkill for such task. I guess you should just revert to the previous file based implementation and request a review again.

Sorry for the hassle.

marcelogdeandrade:
Thanks for the reply! I reverted this branch to my previous file based implementation.

marcelogdeandrade:
@donadigo sorry to bother one more time, I think everything is working right now. Could you review it so I can make any needed changes? Thanks!

src/Views/HardwareView.vala
```diff
@@ -280,6 +282,73 @@ public class About.HardwareView : Gtk.Grid {
...
+        string df_stdout;
+        string df_stderr;
+        string partition = "";
+        int df_status;
+        try {
+            Process.spawn_command_line_sync ("df /",
+                out df_stdout,
+                out df_stderr,
+                out df_status);
+            string[] output = df_stdout.split("\n");
```
donadigo: Space before `(`.

src/Views/HardwareView.vala
```diff
@@ -280,6 +282,73 @@ public class About.HardwareView : Gtk.Grid {
...
+        return storage;
+    }
+
+    private string get_partition_name () {
+        string df_stdout;
+        string df_stderr;
+        string partition = "";
+        int df_status;
+        try {
+            Process.spawn_command_line_sync ("df /",
```
donadigo: You don't need neither of `df_stderr` and `df_status`. You can simply pass `null`'s here.

src/Views/HardwareView.vala
```diff
@@ -280,6 +282,73 @@ public class About.HardwareView : Gtk.Grid {
...
+                out df_stdout,
+                out df_stderr,
+                out df_status);
+            string[] output = df_stdout.split("\n");
+            foreach(string line in output) {
```
donadigo: Space before `(`.

src/Views/HardwareView.vala
```diff
@@ -280,6 +282,73 @@ public class About.HardwareView : Gtk.Grid {
...
+                if (line.has_prefix ("/dev/")) {
+                    string[] partition_line = line.split(" ");
```
donadigo: Space before `(`.

I guess if the line starts with `/dev/` you can safely assume that you can strip everything else after the first space and not bother with a foreach loop.

You could do e.g this:
```
int idx = line.index_of (" ");
if (idx != -1) {
    partition = line.substring (0, idx);
}
```

src/Views/HardwareView.vala
```diff
@@ -280,6 +282,73 @@ public class About.HardwareView : Gtk.Grid {
...
+        string lsblk_stderr;
+        string disk_name = "";
+        int lsblk_status;
+        string command = "lsblk -no pkname " + partition;
+        try {
+            Process.spawn_command_line_sync (command,
+                out lsblk_stout,
+                out lsblk_stderr,
+                out lsblk_status);
+            disk_name = lsblk_stout.strip();
```
donadigo: Space before `(`.

src/Views/HardwareView.vala
```diff
@@ -280,6 +282,73 @@ public class About.HardwareView : Gtk.Grid {
...
+        string lsblk_stout;
+        string lsblk_stderr;
+        string disk_name = "";
+        int lsblk_status;
+        string command = "lsblk -no pkname " + partition;
+        try {
+            Process.spawn_command_line_sync (command,
+                out lsblk_stout,
+                out lsblk_stderr,
+                out lsblk_status);
```
donadigo: Again, you don't need `lslbk_stderr` and `lsblk_status` variables here.

src/Views/HardwareView.vala
```diff
@@ -280,6 +282,73 @@ public class About.HardwareView : Gtk.Grid {
 
         return 0;
     }
+
+    private string get_storage_type () {
+        string partition_name = get_partition_name ();
+        string disk_name = get_disk_name (partition_name);
+        string path = "/sys/block/%s/queue/rotational".printf(disk_name);
+        string storage = "";
+        var rotational_storage_file = File.new_for_path (path);
```
donadigo: You could greatly simplify the code here by doing:
```
try {
    string contents;
    FileUtils.get_contents (path, out contents);
    if (int.parse (contents) == 1) {
        storage = "(SSD)";
    } else {
        storage = "(HDD)";
    }
} catch (FileError e) {
    warning (e.message);
}

File: src/Views/HardwareView.vala [hunk 12]
```diff
@@ -280,6 +282,73 @@ public class About.HardwareView : Gtk.Grid {
 
         return 0;
     }
+
+    private string get_storage_type () {
+        string partition_name = get_partition_name ();
+        string disk_name = get_disk_name (partition_name);
+        string path = "/sys/block/%s/queue/rotational".printf(disk_name);
```
donadigo: Space before `(`.

donadigo:
And I forgot, @danrabbit, should the "(SSD)" and "(HDD)" strings be actually translatable?

marcelogdeandrade:
@donadigo I did the requested changes and also added SSD and HDD strings as translatable! Thanks for your help

src/Views/HardwareView.vala
```diff
@@ -280,6 +282,60 @@ public class About.HardwareView : Gtk.Grid {
...
+        string partition = "";
+        try {
+            Process.spawn_command_line_sync ("df /",
+                out df_stdout);
+            string[] output = df_stdout.split ("\n");
+            foreach (string line in output) {
+                if (line.has_prefix ("/dev/")) {
+                    int idx = line.index_of (" ");
+                    if (idx != -1) {
+                        partition = line.substring (0, idx);
```
donadigo: If you determined the partition, you can break out of the loop here, or even faster, just return it.

marcelogdeandrade: Done :)

jlnr:
Superseded by #93, which has just been merged. Thanks to both of you :muscle:

PR #78, opened 2018-11-29
meson: add option for distributor icon name

codygarver:


decathorpe:
Obvious +1 from me, since this change will allow me to drop the downstream patch in fedora.

danrabbit:
I'm -1 here. This is what the icon `distributor-logo` is for. The distributor is supposed to supply the correct icon in the icon set there

decathorpe:
Just curious, is the `distributor-logo` icon actually documented / standardized anywhere?

danrabbit:
> Just curious, is the `distributor-logo` icon actually documented / standardized anywhere?

That's a good question. I actually can't find any documentation about it, but there are lots of themes that use it. It might just be a "de facto" type thing. I did find a reference [here](https://bugs.launchpad.net/ubuntu/+source/gnome-icon-theme/+bug/91914) for prior art of Ubuntu patching the icon in GNOME icon theme however

danrabbit:
Something I have seen in a couple other themes is something like `distributor-logo-fedora`. I think doing some os detection and appending that to the icon name (with fallbacks) would probably better support custom icon themes and still allow you to supply your own icon assets as a distribution

decathorpe:
I just checked, and the only icon themes providing a `distributor-logo-fedora` icon on fedora are: Paper, Numix, papirus, Moka, Pop, xenlism-wildfire (seems like this is something @snwh likes to do in his themes).

So, computing this icon name at runtime would only work when using one of those themes. Neither elementary, hicolor, or GNOME (default on fedora) icon themes contain such an icon - I don't think that's a workable approach.

Since there's no standard way to do this, I think a build-time argument is the easiest, and most flexible solution: The person who builds the package also has the necessary knowledge of the correct icon name, and no failure-prone heuristics are necessary at runtime.

codygarver:
The argument against this change, that the active theme should be supplying this icon, is the problem this change is trying to solve.

Using the elementary icon theme as an example, it's shipping an elementary/elementary OS branded `distributor-logo` which doesn't make sense to display outside of elementary OS.

This gets even more complicated when a distro is based on another distro.

For example, even if it was an openSUSE branded `distributor-logo` in the theme, it would still be causing problems on a distro that carries unique branding.

Without this PR, the only recourse is to fork and remix the whole default icon theme just to change the `distributor-logo` to match branding.

codygarver:
It's also worth noting here that Fedora and openSUSE ship things like this in a dedicated package and install them to `hicolor` etc to avoid encroaching on branding, unlike the intentions of the named icon themes.

codygarver:
Superseded by #79

danrabbit:
@codygarver I agree we should pull `distributor-logo` out of the icon set. There's an issue report about it [here](https://github.com/elementary/icons/issues/377). Can you comment on that thread if the proposed solution there makes sense or not?

Issue #85, opened 2019-01-04
Epiphany missing in Report a problem dialog


peteruithoven:
The Epiphany browser is missing in the *Report a problem* dialog.
![screenshot from 2019-01-04 17 37 03](https://user-images.githubusercontent.com/523210/50699208-78fbc880-1047-11e9-97eb-2edbd0d125e4.png)

It should probably link to https://gitlab.gnome.org/GNOME/epiphany/ ?

danrabbit:
I think we might want some feedback from @mcatanzaro about whether he wants Epiphany issues from elementary OS users in GitLab or if he would prefer them to go to a separate tracker since we have some downstream patches

mcatanzaro:
Hm, as long as you have a recent version, reporting upstream is fine. I don't mind redirecting users to the downstream bugtracker whenever an elementary patch is implicated. That shouldn't happen very often.

One complication is that we don't accept web content bug reports on the Epiphany issue tracker; these should be reported to WebKit Bugzilla instead. How does the problem reporting app work? Does it just link to the upstream bugtracker? If so, you could link to https://gitlab.gnome.org/GNOME/epiphany/blob/master/CONTRIBUTING.md I suppose, something like that?

Issue #69, opened 2018-09-21
Add ability to attach basic system info with report a issue dialog


arshubham:
I would like attach some basic info/possibly more detailed too (Option disabled by default) about the hardware while reporting issues which would help the developers solve the issues faster by giving them more info.
This should contain purely technical info with no user identifiable info about installed software or IP address.

peteruithoven:
How would this work though?
The current Report a problem functionality seems to redirect the user to the right issue queue, which is probably smart to stimulate searching to existing issues first.

I've opened the following issue in the past with a similar goal:
https://github.com/elementary/switchboard-plug-about/issues/47

arshubham:
@peteruithoven I have not checked but there might be a way to send api request with github with the hw info as payload, which will be appended to the issue. And since the dialog is only for first party apps, devs can control what info they collect for a repo. For sound related apps they in advaced info they should collect pulseaudio logs and related info or in case of gala issues they should collect driver logs (nvidia/mesa) and versions of those packages involved.

PR #86, opened 2019-01-20
Added support for Epiphany in Report a Problem

KeironO:
This pull request aims to resolve https://github.com/elementary/switchboard-plug-about/issues/85.

### DIFF
--- a/src/Dialogs/IssueDialog.vala
+++ b/src/Dialogs/IssueDialog.vala
@@ -63,22 +63,22 @@ public class About.IssueDialog : Granite.MessageDialog {
 
         foreach (var app in app_entries) {
             var desktop_info = new DesktopAppInfo (app.app_id + ".desktop");
-            var repo_row = new RepoRow (desktop_info.get_display_name (), desktop_info.get_icon (), Category.DEFAULT_APPS, app.github_suffix);
+            var repo_row = new RepoRow (desktop_info.get_display_name (), desktop_info.get_icon (), Category.DEFAULT_APPS, app.issues_url);
             listbox.add (repo_row);
         }
 
         foreach (var entry in system_entries) {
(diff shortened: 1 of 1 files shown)


cassidyjames:
@danrabbit thoughts on this? It makes sense to me but does add a little but of noise, but reduces cleverness. Which is probably fine?

cassidyjames:
An alternative would be to set a GitHub string const if we wanted to reduce repetition there, but that also just adds some complexity. Maybe specifying an entire URL is fine.

arshubham:
@KeironO Can a identifier be added to see if issue is a github/gitlab or Other url. This will help to solve https://github.com/elementary/switchboard-plug-about/issues/69 . Based on type of url github/gitlab api can be used to send user info.

KeironO:
> @KeironO Can a identifier be added to see if issue is a github/gitlab or Other url. This will help to solve #69 . Based on type of url github/gitlab api can be used to send user info.

Hey @arshubham,

I have been unable to find any documentation from either GitHub or Gitlab that outlines that such functionality exists. If you can direct me to it, I would have no issue in attempting to resolve said issue.

arshubham:
Can't this endpoint be used https://developer.github.com/v3/issues/#create-an-issue ? Just send an empty string as a title and body with the user info.

arshubham:
Can't this endpoint be used https://developer.github.com/v3/issues/#create-an-issue ? Just send an empty string as a title and body with the user info.

peteruithoven:
I would focus this PR on adding Epiphany (gitlab) support. Maybe a separate PR could be opened to work on #69? 
@arshubham maybe you could add those tips to #69?

KeironO:
I'll get to work on this this weekend.

PR #91, opened 2019-06-21
Launch Feedback App

danrabbit:
When clicking "Report a Problem", tries to launch the feedback app. Otherwise, launch the support URL


src/Plug.vala
```diff
@@ -197,12 +197,22 @@ public class About.Plug : Switchboard.Plug {
...
         bug_button.clicked.connect (() => {
-            var issue_dialog = new IssueDialog ();
-            issue_dialog.transient_for = (Gtk.Window) main_grid.get_toplevel ();
-            issue_dialog.run ();
+            var appinfo = new GLib.DesktopAppInfo  ("io.elementary.feedback.desktop");
+            if (appinfo != null) {
+                try {
+                    appinfo.launch (null, null);
+                } catch (Error e) {
+                    critical (e.message);
```
donadigo: `launch_default_for_uri` should happen here too.

src/Plug.vala
```diff
@@ -197,12 +197,22 @@ public class About.Plug : Switchboard.Plug {
             }
         });
 
-        // Bug button
         var bug_button = new Gtk.Button.with_label (_("Report a Problem"));
         bug_button.clicked.connect (() => {
-            var issue_dialog = new IssueDialog ();
-            issue_dialog.transient_for = (Gtk.Window) main_grid.get_toplevel ();
-            issue_dialog.run ();
+            var appinfo = new GLib.DesktopAppInfo  ("io.elementary.feedback.desktop");
```
ryonakano:
```suggestion
            var appinfo = new GLib.DesktopAppInfo ("io.elementary.feedback.desktop");
```

This might be a nitpicking, but this is double-spaced.

PR #93, opened 2019-06-29
Show what kind of storage is used

### LINKED ISSUE #63: Show what kind of storage is used

ryonakano:
Fixes #63
Supersedes #67

I made this PR because it looks like the branch of that PR no longer exists :cry:
The original PR checks whether system storage is SSD or HDD, but this PR also recognizes whether it's NVMe, eMMC, or SATA.

## Changes Summary

* Show whether the system storage is NVMe SSD / eMMC / SATA SSD / HDD (#67 + α)

### With NVMe SSD

Screenshot from 2019-06-29 20-53-33

### With SATA SSD

Screenshot from 2019-06-29 20-57-09

### With SATA HDD

Screenshot from 2019-06-29 21-02-23

### DIFF
--- a/src/Views/HardwareView.vala
+++ b/src/Views/HardwareView.vala
@@ -294,7 +294,13 @@ public class About.HardwareView : Gtk.Grid {
             string contents;
             FileUtils.get_contents (path, out contents);
             if (int.parse (contents) == 0) {
-                storage = _("SSD");
+                if (disk_name.has_prefix ("nvme")) {
+                    storage = _("NVMe SSD");
+                } else if (disk_name.has_prefix ("mmc")) {
+                    storage = _("eMMC");
+                } else {
+                    storage = _("SATA SSD");
+                }
             } else {
                 storage = _("HDD");
             }
--- a/src/Views/HardwareView.vala
+++ b/src/Views/HardwareView.vala
@@ -293,14 +293,14 @@ public class About.HardwareView : Gtk.Grid {
             FileUtils.get_contents (path, out contents);
             if (int.parse (contents) == 0) {
                 if (disk_name.has_prefix ("nvme")) {
-                    storage = _("NVMe SSD");
+                    storage = _("storage (NVMe SSD)");
                 } else if (disk_name.has_prefix ("mmc")) {
-                    storage = _("eMMC");
+                    storage = _("storage (eMMC)");
                 } else {
-                    storage = _("SATA SSD");
+                    storage = _("storage (SATA SSD)");
                 }
             } else {
-                storage = _("HDD");
+                storage = _("storage (HDD)");
             }
         } catch (FileError e) {
             warning (e.message);
--- a/src/Views/HardwareView.vala
+++ b/src/Views/HardwareView.vala
@@ -64,7 +64,7 @@ public class About.HardwareView : Gtk.Grid {
         graphics_info.justify = Gtk.Justification.CENTER;
         graphics_info.set_selectable (true);
 
-        ///TRANSLATORS: The first %s represents capacity of system strage.
+        ///TRANSLATORS: The first %s represents capacity of system storage.
         ///The second one is its type: whether HDD or SSD.
         var hdd_info = new Gtk.Label (_("%s %s").printf (hdd, get_storage_type ()));
         hdd_info.ellipsize = Pango.EllipsizeMode.END;
--- a/src/Views/HardwareView.vala
+++ b/src/Views/HardwareView.vala
@@ -65,7 +65,9 @@ public class About.HardwareView : Gtk.Grid {
         graphics_info.justify = Gtk.Justification.CENTER;
         graphics_info.set_selectable (true);
 
-        var hdd_info = new Gtk.Label (_("%s storage %s").printf (hdd, storage_type));
+        ///TRANSLATORS: The first %s represents capacity of system strage.
+        ///The second one is its type: whether HDD or SSD.
+        var hdd_info = new Gtk.Label (_("%s %s").printf (hdd, storage_type));
         hdd_info.ellipsize = Pango.EllipsizeMode.END;
         hdd_info.set_selectable (true);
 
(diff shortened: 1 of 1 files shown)


src/Views/HardwareView.vala
```diff
@@ -280,6 +282,68 @@ public class About.HardwareView : Gtk.Grid {
...
+                    storage = _("eMMC");
+                } else {
+                    storage = _("SATA SSD");
+                }
+            } else {
+                storage = _("HDD");
+            }
+        } catch (FileError e) {
+            warning (e.message);
+            storage = _("Unknown");
```
vjr: Instead of "unknown" why not say "storage" like a fallback to the existing output?

src/Views/HardwareView.vala
```diff
@@ -64,7 +64,9 @@ public class About.HardwareView : Gtk.Grid {
         graphics_info.justify = Gtk.Justification.CENTER;
         graphics_info.set_selectable (true);
 
-        var hdd_info = new Gtk.Label (_("%s storage").printf (hdd));
+        ///TRANSLATORS: The first %s represents capacity of system storage.
+        ///The second one is its type: NVMe SSD / eMMC / SATA SSD / HDD.
+        var hdd_info = new Gtk.Label (_("%s %s").printf (hdd, get_storage_type ()));
```
jlnr: Can we use position placeholders here? Otherwise translators won't be able to swap the order if necessary.

src/Views/HardwareView.vala
ryonakano: Thank you, addressed.

src/Views/HardwareView.vala
ryonakano: Fixed. Thank you!

src/Views/HardwareView.vala
```diff
@@ -280,6 +282,68 @@ public class About.HardwareView : Gtk.Grid {
...
+                    storage = _("eMMC");
+                } else {
+                    storage = _("SATA SSD");
+                }
+            } else {
+                storage = _("HDD");
+            }
+        } catch (FileError e) {
+            warning (e.message);
+            storage = _("Storage");
```
jlnr: It was lowercase before, so I think we should keep it for consistency. Maybe also worth adding a comment to indicate that this is just the fallback string.

ryonakano:
> The code correctly detects my disk, but "240 GB SATA SSD" is less self-explanatory than "240 GB storage".

In the previous PR (#67), it looks like "240 GB storage (SSD)", but I felt saying "storage" is redundant because the "D" of "SSD" or "HDD" means "disk" which is the essentially the same meaning with "storage", and removed it. Indeed, however, removing "storage" is less self-explanatory, especially for those who's not used to computers. Maybe "240 GB storage (SATA SSD)" is even better?

src/Views/HardwareView.vala
ryonakano: Addressed

jlnr:
> Maybe "240 GB storage (SATA SSD)" is even better?

I think that's better and there's enough space for this information.

ryonakano:
Then should we change the following string to "Unknown" again, because if an FileError occurred during parings, storage label will look like "240 GB storage (storage)"?

https://github.com/ryonakano/switchboard-plug-about/blob/84151e3a0db5da26dcf39616145caea9bee06264/src/Views/HardwareView.vala#L308

vjr:
> Then should we change the following string to "Unknown" again, because if an FileError occurred during parings, storage label will look like "240 GB storage (storage)"?
> 
> https://github.com/ryonakano/switchboard-plug-about/blob/84151e3a0db5da26dcf39616145caea9bee06264/src/Views/HardwareView.vala#L308

If it's unknown then just "240 GB storage" should be shown I guess without the parentheses.

src/Views/HardwareView.vala
```diff
@@ -280,6 +284,69 @@ public class About.HardwareView : Gtk.Grid {
...
+        string partition_name = get_partition_name ();
+        string disk_name = get_disk_name (partition_name);
+        string path = "/sys/block/%s/queue/rotational".printf (disk_name);
+        string storage = "";
+        try {
+            string contents;
+            FileUtils.get_contents (path, out contents);
+            if (int.parse (contents) == 0) {
+                if (disk_name.has_prefix ("nvme")) {
+                    storage = _("storage (NVMe SSD)");
```
jlnr: Sorry for going in circles here :sweat_smile: But at this point, wouldn't it be easier to have individual strings like "%s storage (NVMe SSD)", etc. for each storage type?

ryonakano: Fixed!

jlnr:
Not sure why there are two Travis checks and one of them is still pending. How to trigger a rebuild? With a new dummy commit maybe?

ryonakano:
コメント 2019-07-06 225703

> Not sure why there are two Travis checks and one of them is still pending. How to trigger a rebuild? With a new dummy commit maybe?

Hmm as far as I understand, the "Pull Request" (or "pr") and "Branch" check are executed if we make a PR from a branch in the same repository, but if we make a PR from a branch in a forked repository the latter check should not be done like the screenshot above. However, for some reason the latter check is also displayed as "pending" in this PR. I'm not sure either why or how to fix, but I think it's OK because the former check is finished successfully.

PR #96, opened 2019-07-30
Remove arch_name

danrabbit:
Since we don't provide 32bit images anymore and we've never provided ARM images, do we need this?


donadigo:
Even though we don't provide these images, this info could be useful for other distros that use this plug in some way. I'm not sure why would we really need to remove this.

cassidyjames:
Probably want to hear from Fedora folks or other downstreams if this is useful? I'm personally ambivalent.

worldofpeace:
Downstream from NixOS, I think cpu architecture is important information to display regardless as to whether an OS supports it. Additionally NixOS formally supports Aarch64.  

Why isn't this info in the HardwareView?

cassidyjames:
@worldofpeace I _think_ because it's the software version, not the hardware architecture. So if you installed a 32-bit OS on 64-bit hardware, this would say 32-bit. If that's not the case, then yeah we should probably move it to hardware.

worldofpeace:
> @worldofpeace I _think_ because it's the software version, not the hardware architecture. So if you installed a 32-bit OS on 64-bit hardware, this would say 32-bit. If that's not the case, then yeah we should probably move it to hardware.

If that's the case that's a very fishy topic, because in general binaries aren't portable between CPU architectures. Especially x86 and ARM (not to mention its many flavors).

cassidyjames:
@worldofpeace I would need to test this to see if that's true. But I know in the past we offered 32- and 64-bit builds of elementary OS, and you can indeed run 32-bit on 64-bit, or arm32 on arm64 (but not the other way around).

Conan-Kudo:
From the Fedora perspective, I'd say showing the CPU arch makes sense, but saying `32-bit`/`64-bit` definitely does not.

PR #97, opened 2019-08-05
build: make update button optional

worldofpeace:
In environments that don't have Appcenter
or will never have a use for it this button
would be confusing.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -13,6 +13,10 @@ add_project_arguments(
     language:'c'
 )
 
+if get_option('with-update-button')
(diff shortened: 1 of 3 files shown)


worldofpeace:
Huh, I thought a compile time option was fine.
Checking if it's installed with `GLib.Environment.find_program_in_path` might be less intrusive since you don't have to do any build configuration. I'm not sure I'd make that change though.

danrabbit:
superseded by #98

worldofpeace:
> superseded by #98

:sparkling_heart: Another downstream patch gone.

PR #98, opened 2019-08-05
Plug: check for AppCenter before creating updates button

danrabbit:
Alternative to #97 that checks for the existence of AppCenter before creating the update button

### DIFF
--- a/src/Plug.vala
+++ b/src/Plug.vala
@@ -212,15 +212,19 @@ public class About.Plug : Switchboard.Plug {
             }
         });
 
-        // Update button
-        var update_button = new Gtk.Button.with_label (_("Check for Updates"));
-        update_button.clicked.connect (() => {
-            try {
-                Process.spawn_command_line_async ("io.elementary.appcenter --show-updates");
-            } catch (Error e) {
-                warning (e.message);
-            }
-        });
+        Gtk.Button? update_button = null;
+        var appcenter_info = new GLib.DesktopAppInfo ("io.elementary.appcenter");
+        if (appcenter_info != null) {
+            update_button = new Gtk.Button.with_label (_("Check for Updates"));
+            update_button.clicked.connect (() => {
+                try {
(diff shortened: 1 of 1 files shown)


cassidyjames:
@danrabbit for some reason I'm still not getting an updates button after reinstalling AppCenter from the repos, rebooting, etc.

src/Plug.vala
```diff
@@ -212,15 +212,19 @@ public class About.Plug : Switchboard.Plug {
...
-                warning (e.message);
-            }
-        });
+        Gtk.Button? update_button = null;
+        var appcenter_info = new GLib.DesktopAppInfo ("io.elementary.appcenter");
+        if (appcenter_info != null) {
+            update_button = new Gtk.Button.with_label (_("Check for Updates"));
+            update_button.clicked.connect (() => {
+                try {
+                    Process.spawn_command_line_async ("io.elementary.appcenter --show-updates");
```
tintou:
```suggestion
                    appcenter_info.launch_action ("ShowUpdates", null);
```
That way it looks better :)

src/Plug.vala
```diff
@@ -213,15 +213,11 @@ public class About.Plug : Switchboard.Plug {
...
+        var appcenter_info = new GLib.DesktopAppInfo ("io.elementary.appcenter.desktop");
         if (appcenter_info != null) {
             update_button = new Gtk.Button.with_label (_("Check for Updates"));
             update_button.clicked.connect (() => {
-                try {
-                    Process.spawn_command_line_async ("io.elementary.appcenter --show-updates");
-                } catch (Error e) {
-                    warning (e.message);
-                }
+                appcenter_info.launch_action ("ShowUpdates", new GLib.AppLaunchContext ());
```
cassidyjames: TIL that's a thing 🤯️

danrabbit: Is that an approve :p

PR #101, opened 2019-09-13
Plug: Remove updates from search results

danrabbit:
Searching in the Applications menu gives you:

Screenshot from 2019-09-12 17 16 19@2x

This branch gives you:

Screenshot from 2019-09-12 17 21 32@2x

Less is more in this case


donadigo:
Won't that basically remove the search term also from Switchboard? I imagine that this could be useful when someone searches in settings.

jeremypw:
I can confirm that this PR currently stops searching for "Updates" in Switchboard stops anything being returned.

PR #133, opened 2020-07-23
Support ARM CPUs and use libgtop to get info

davidmhewitt:
Instead of parsing `/proc/cpuinfo` and `/proc/meminfo` manually, we use a couple of methods in GTop to do this. This means it should work a little better on platforms where those files are a bit more locked down and the code is slightly cleaner.

ARM CPUs just report a couple of IDs in hex to identify the model rather than a string, so we need some lookup tables for those.

I also took the liberty to add "Hexa-Core" as a string while I was here since those are becoming more common now (I have an AMD one!)


davidmhewitt:
I checked on my Intel-based laptop, but maybe worth checking on a couple more x86 based machines that the CPU and RAM info is still correct. Especially if we have any AMD based machines anywhere.

Issue #90, opened 2019-05-04
Make hostname editable


vjr:
Related to the discussion in https://github.com/elementary/installer/issues/383 this is a wishlist to make the hostname field configurable, similar to what is available in other OSes like Fedora for example.

arshubham:
We might want to take a look at [hostnamectl](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/networking_guide/sec_configuring_host_names_using_hostnamectl) . It has options to set --pretty, --static, and --transient hostnames.

cassidyjames:
GNOME Control Center has a way to edit, we should take a look at that as well.

PR #139, opened 2020-09-17
hostname with Gtk.sizegroup

torikulhabib:
this way seems better.
Screen record from 2020-09-17 23 29 19

### DIFF
--- a/src/Views/HardwareView.vala
+++ b/src/Views/HardwareView.vala
@@ -33,6 +33,8 @@ public class About.HardwareView : Gtk.Grid {
     private string product_version;
     private SystemInterface system_interface;
     private SessionManager session_manager;
+    private Gtk.Label product_name_info;
+    private Gtk.Entry hostname_entry;
 
     construct {
         try {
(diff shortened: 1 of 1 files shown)


vjr:
Fixes #90 

(i have not tried this branch myself though)

cassidyjames (changes requested):
Thanks for working on this, and thanks for your patience in waiting for a review! UX-wise I have at least one change and some concerns:

- [ ] This is not discoverable. I think an edit icon next to the hostname might work, or just a more explicit labeled button.
- [ ] How does this interact with OEM-provided information, i.e. a model name instead of the hostname?

I almost wonder if this would better belong in the Sharing plug? I recall it being in a place like that on Android, at least. And then we could link to the sharing plug to edit your hostname from the About plug, Bluetooth plug, etc. wherever someone might expect to be able to change it.

Thoughts from other @elementary/ux folks?

danrabbit:
@cassidyjames Yeah I agree sharing seems appropriate here. There's a lot more room and like you said we can link from other places. Since there hasn't been any response to this since November, I'm inclined to close this PR

PR #141, opened 2020-10-21
Updates to use switcheroo-control to get dual GPU information

jhaygood86:
This switches to using switcheroo-control to get dual GPU information.

Note: switcheroo-control is already packaged by Ubuntu, but not installed by default on elementary OS 6.0. This package will need to depend on that. I am also going to work on updating the Application Menu to use the package details to support the Launch on Dedicated/Integrated Card feature like GNOME 3.38 has (which uses a setting in the .desktop file to determine if it should default to the Integrated or Dedicated card, and uses a context menu to give the user to option use a different GPU)

To be fully useful on NVIDIA systems, we will need to make sure NVIDIA proprietary kernel is installed AND prime-select is set to on-demand (i.e., use hybrid graphics)


jhaygood86:
Screenshot from 2020-10-21 00-07-29

Example screenshot on a Hybrid Graphics system

andirsun:
My nvidia mx 250 graphic card doesnt show properly name. Will this PR fix that ? 
image

README.md
```diff
@@ -13,6 +13,7 @@ You'll need the following dependencies:
 * libgtop2-dev
 * meson
 * valac
+* switchboard-control (at runtime)
```
davidmhewitt, line 16: You mean `switcheroo-control` here, right?

davidmhewitt (changes requested):
Looks good at a glance!

I've added one minor comment about the README.

I don't have any dual GPU machines to test. But I'll test on a few of my single GPU machines to check there are no regressions and get back to you.

davidmhewitt (changes requested):
Additionally, I would prefer if we could keep the gnome-session code around as a fallback for getting the GPU info if switcheroo isn't working/installed.

src/Views/HardwareView.vala
```diff
@@ -267,6 +279,23 @@ public class About.HardwareView : Gtk.Grid {
         }
     }
 
+    private string get_gpu_name (bool is_primary) {
```
davidmhewitt, line 282: The return type of this method needs to be marked as nullable with `string?` as returning `null` is a possibility with this method.

davidmhewitt (changes requested):

davidmhewitt:
@jhaygood86 Do you still have time to work on this? I'd be happy to take over and finish this branch off if you like?

davidmhewitt:
I reworked this branch and made some improvements and it has now been merged in https://github.com/elementary/switchboard-plug-about/commit/35102bb4189d34e2bd6b81485e8482dbc1fdc476

So I'll close this PR now. Thanks again for starting this, it's a big improvement!

PR #146, opened 2020-11-15
Rename user-facing strings to System

### LINKED ISSUE #124: Rename to "System"?
This plug provides some info about the system, but it also holds a handful of actions: issue reporting, help, checking for updates, links to websites, and soon (#123) maybe firmware updates and eventually OS upgrades.

So maybe it makes sense to rename to "System" or something similar instead of "About"?

cassidyjames:
Fixes #124, at least in user-facing strings


meisenzahl (approved):
Looks good to me.

PR #149, opened 2020-11-25
Fix plug destroyed on navigation

### LINKED ISSUE #145: Plug appears to be destroyed when navigating away
## What Happened

When you hit "All Settings" to leave the About plug, the plug contents themselves seem to disappear. This is made more obvious since https://github.com/elementary/switchboard/pull/181, where it also breaks the animation.

## Expected Behavior

Navigating away from the About plug is smooth and matches the other plugs.

## Steps to Reproduce

1. Open the About plug
2. Hit "All Settings" to leave it

## Platform Information

elementary OS 6 daily, running Switchboard master from Git.

pongloongyeat:
Fixes #145

### DIFF
--- a/src/Plug.vala
+++ b/src/Plug.vala
@@ -48,7 +48,6 @@ public class About.Plug : Switchboard.Plug {
     }
 
     public override void hidden () {
-        main_grid.hide ();
     }
 
     public override void search_callback (string location) {


danrabbit (approved):
Huh yeah that seems unnecessary. Nice catch!

Issue #123, opened 2017-10-09
Add support for fwupd


ma1onso:
Example: 
![h3wxye1](https://user-images.githubusercontent.com/1877511/31346868-22bb91d8-ace0-11e7-95e6-7fdcdc035749.png)

https://fwupd.org/

PR #150, opened 2020-12-10
Add support for fwupd

meisenzahl:
Closes https://github.com/elementary/switchboard-plug-about/issues/123


src/Interfaces/FwupdManager.vala
```diff
@@ -0,0 +1,124 @@
...
+                            break;
+                        case "Vendor":
+                            device.vendor = _device.lookup (key).get_string ();
+                            break;
+                        case "Version":
+                            device.version = _device.lookup (key).get_string ();
+                            break;
+                        case "Icon":
+                            var icons = _device.lookup (key).get_strv ();
+                            device.icon = icons.length > 0 ? icons[0] : "unknown";
```
danrabbit, line 76: We should probably fall back to `application-x-firmware`

danrabbit (commented):
Something I noticed is that this introduces a significant lag when first opening the page. There are probably some operations that need to be made asynchronous from the view construction

meisenzahl:
> Something I noticed is that this introduces a significant lag when first opening the page. There are probably some operations that need to be made asynchronous from the view construction

@danrabbit I added an interface to interact with fwupd via D-Bus. I'm afraid that by default the method calls are synchronous. Are you aware of a repository where asynchronous calls are used?

meisenzahl:
## Current state

This is now a minimal working prototype. It lists only the latest available update per device.

If an older version of the firmware is flashed, you can click **Update**. The firmware file is downloaded and installed via fwupd D-Bus service.

For now the interaction with fwupd via D-Bus is synchronous.

Peek 2020-12-27 16-17

meisenzahl:
## Current state

The D-Bus methods calls are now asynchronous.

The UI is pretty minimal. Feedback on interactions for the user is completely missing. I am looking forward to suggestions 🙂

Peek 2020-12-30 07-12

meisenzahl:
## Current state

A visual feedback is now provided if a device is being updated.

Peek 2021-01-03 05-09

src/Interfaces/FwupdManager.vala
```diff
@@ -0,0 +1,294 @@
...
+                debug ("%" + int64.FORMAT + " bytes of %" + int64.FORMAT + " bytes copied.",
+                    current_num_bytes, release.size);
+            });
+        } catch (Error e) {
+            warning ("Could not download file: %s", e.message);
+            return;
+        }
+
+        if (!result) {
+            warning ("Download of %s was not succesfull", release.uri);
```
danrabbit, line 243: These should probably be user-facing error dialogs right? Otherwise this would just fail silently

src/Plug.vala
```diff
@@ -36,15 +33,34 @@ public class About.Plug : Switchboard.Plug {
...
+            stack.add_titled (hardware_view, "hardware", _("Hardware"));
+
+            var firmware_view = new FirmwareView ();
+            stack.add_titled (firmware_view, "firmware", _("Firmware"));
+
+            var stack_switcher = new Gtk.StackSwitcher ();
+            stack_switcher.stack = stack;
+            stack_switcher.halign = Gtk.Align.CENTER;
+            stack_switcher.homogeneous = true;
+            stack_switcher.margin = 24;
```
danrabbit, line 53: This should use the latest code style:
```suggestion
            var stack_switcher = new Gtk.StackSwitcher () {
                halign = Gtk.Align.CENTER,
                homogeneous = true,
                margin = 24,
                stack = stack
            };
```

src/Plug.vala
```diff
@@ -36,15 +33,34 @@ public class About.Plug : Switchboard.Plug {
...
+
+            main_grid.attach (stack_switcher, 0, 0, 1, 1);
+            main_grid.attach (stack, 0, 1, 1, 1);
```
danrabbit, line 56: attach uses a width and height of 1 by default so this can be:
```suggestion
            main_grid.attach (stack_switcher, 0, 0);
            main_grid.attach (stack, 0, 1);
```

src/Plug.vala
```diff
@@ -16,11 +16,8 @@
 //
 
 public class About.Plug : Switchboard.Plug {
-    private string kernel_version;
-    private string support_url;
-    private Gtk.Label based_off;
-
-    private string upstream_release;
+    private Gtk.Stack stack;
+    private OperatingSystemView operating_system_view;
```
danrabbit, line 20: Is there a reason these are scoped to the plug class? They don't seem to be accessed outside of `get_widget ()`

src/Views/FirmwareView.vala
```diff
@@ -0,0 +1,117 @@
...
+*
+* You should have received a copy of the GNU General Public
+* License along with this program; if not, write to the
+* Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+* Boston, MA 02110-1301 USA
+*
+* Authored by: Marius Meisenzahl <mariusmeisenzahl@gmail.com>
+*/
+
+public class About.FirmwareView : Granite.SettingsPage {
```
danrabbit, line 22: is there a reason not to use `SimpleSettingsPage`? That would handle all the title area creation for you

src/Views/FirmwareView.vala
```diff
@@ -0,0 +1,117 @@
...
+
+        grid.attach (header_icon, 0, 0);
+        grid.attach (title_label, 1, 0);
+        grid.attach (frame, 0, 1, 3, 1);
+
+        stack = new Gtk.Stack ();
+        stack.transition_type = Gtk.StackTransitionType.SLIDE_LEFT_RIGHT;
+
+        stack.add (grid);
+        stack.add (progress_view);
```
danrabbit, line 86: You could use latest code style here:
```suggestion
        stack = new Gtk.Stack () {
            transition_type = Gtk.StackTransitionType.SLIDE_LEFT_RIGHT
        };
        stack.add (grid);
        stack.add (progress_view);
```

src/Views/FirmwareView.vala
```diff
@@ -0,0 +1,117 @@
...
+
+        var scrolled_window = new Gtk.ScrolledWindow (null, null);
+        scrolled_window.add (update_list);
+
+        var frame = new Gtk.Frame (null);
+        frame.add (scrolled_window);
+
+        grid.attach (header_icon, 0, 0);
+        grid.attach (title_label, 1, 0);
+        grid.attach (frame, 0, 1, 3, 1);
```
danrabbit, line 80: Please group like items so that things stay organized and legible

```suggestion
        update_list = new Gtk.ListBox () {
            vexpand = true,
            selection_mode = Gtk.SelectionMode.SINGLE
        };

        var scrolled_window = new Gtk.ScrolledWindow (null, null);
        scrolled_window.add (update_list);

        var frame = new Gtk.Frame (null);
        frame.add (scrolled_window);

        grid = new Gtk.Grid () {
            column_spacing = 12,
            row_spacing = 12,
            margin = 12
        };
        grid.attach (header_icon, 0, 0);
        grid.attach (title_label, 1, 0);
        grid.attach (frame, 0, 1, 3, 1);
```

src/Views/FirmwareView.vala
```diff
@@ -0,0 +1,117 @@
...
+        stack.add (grid);
+        stack.add (progress_view);
+
+        add (stack);
+
+        update_list_view.begin ();
+    }
+
+    private async void update_list_view () {
+        foreach (Gtk.Widget element in update_list.get_children ()) {
```
danrabbit, line 94: This should be an unowned reference:
```suggestion
        foreach (unowned Gtk.Widget element in update_list.get_children ()) {
```

It would probably also be better to use `widget` as a variable name instead of `element` since this is a `Gtk.Widget`

src/Views/FirmwareView.vala
```diff
@@ -0,0 +1,117 @@
...
+            if (element is Gtk.ListBoxRow) {
+                update_list.remove (element);
+            }
+        }
+
+        foreach (var device in yield FwupdManager.get_instance ().get_devices ()) {
+            if (device.is (DeviceFlag.UPDATABLE) && device.releases.length () > 0) {
```
danrabbit, line 101: Shouldn't we also show devices that are up to date? If we filter to only show devices with updates, there's no way to see that a device is in fact running the latest firmware version

src/Widgets/FirmwareUpdateWidget.vala
```diff
@@ -0,0 +1,91 @@
...
+public class About.Widgets.FirmwareUpdateWidget : Gtk.ListBoxRow {
+    public Device device { get; construct set; }
+
+    public signal void on_update_start ();
+    public signal void on_update_end ();
+
+    public FirmwareUpdateWidget (Device device) {
+        GLib.Object (
+            device: device
+        );
```
danrabbit, line 31: This can be simplified a bit to:
```suggestion
        Object (device: device);
```

src/Widgets/FirmwareUpdateWidget.vala
```diff
@@ -0,0 +1,91 @@
...
+        device_name_label.get_style_context ().add_class (Granite.STYLE_CLASS_H3_LABEL);
+
+        var version_label = new Gtk.Label (device.latest_release.version) {
+            wrap = true,
+            xalign = 0
+        };
+
+        var grid = new Gtk.Grid ();
+        grid.column_spacing = 12;
+        grid.margin = 6;
```
danrabbit, line 52: Code style:
```suggestion
        var grid = new Gtk.Grid () {
            column_spacing = 12,
            margin = 6
        };
```

danrabbit (changes requested):
Thanks for all of your effort on this! Everything loads instantly now. Nice work!

Just a quick preliminary pass. Left some comments about some code style issues and a couple of questions. It would be nice to also have someone like @davidmhewitt or @tintou review some of the more complex stuff

Something I noticed is that there is no placeholder in the device list in case there are no available devices. We should probably show some kind of messaging indicating that we're unable to provide updates for device firmware at this time. I'm not sure what else could be helpful here. Maybe a link to fwupd.org for more information? Maybe @cassidyjames has copy ideas here

I left some comments about silently failing when installing firmware. We want to make sure that we're providing feedback to users when things go wrong

src/Views/FirmwareView.vala
meisenzahl: This filter is for devices which are not capable of being updated.

meisenzahl:
## Current state

Show placeholder when no devices are available.

Bildschirmfoto von 2021-01-13 19 39 38

meisenzahl:
## Current state

Show dialog if error occurs during update

Bildschirmfoto von 2021-01-13 20 09 51

@danrabbit is it possible to get the window in a plug? Otherwise I'm not able to set `transient_for` of the dialog.

src/Views/FirmwareView.vala
```diff
@@ -0,0 +1,157 @@
...
+        progress_view = new Gtk.Grid () {
+            margin = 24
+        };
+        progress_view.attach (progress_alert_view, 0, 0);
+
+        var no_devices_alert_view = new Granite.Widgets.AlertView (
+            _("No devices available"),
+            _("We're unable to provide updates for device firmware."),
+            "application-x-firmware"
+        );
```
cassidyjames, line 53: copy tweak

```suggestion
        var no_devices_alert_view = new Granite.Widgets.AlertView (
            _("No devices available"),
            _("Firmware updates are not supported on this or any connected devices."),
            "application-x-firmware"
        );
```

cassidyjames (changes requested):
Just a tiny thing. We (ha) don't refer to the system as "we" in copy, but should rephrase it to something like this.

src/Views/FirmwareView.vala
cassidyjames: @elementary/ux is there a better phrase here than "updatable"? It's not that there aren't any devices, or that there's no firmware, it's that there aren't and devices connected that are able to be updated.

meisenzahl: Thanks for your feedback. I'm always impressed how understandable you express things with few words 👀

lainsce: Maybe "No devices able to update"?

danrabbit:
@meisenzahl yes, you should be able to do `(Gtk.Window) get_toplevel`

src/Views/FirmwareView.vala
cassidyjames: I'm actually wondering if we even need the title and icon… we have the tab right above with "Firmware", and this view looks fine with just a list of devices.

Screenshot from 2021-01-14 13 28 41

cassidyjames:
@meisenzahl On my Star LabTop which supports "system" firmware updates, if I try to update I get this screen, then it goes away without any indication of success or error.

Screenshot from 2021-01-14 13-32-14

GNOME Firmware installs it, then shows this:

Screenshot from 2021-01-14 13-34-31

So I guess we need to handle this?

meisenzahl:
## Current state

- Show dialog with instructions for update
- Show dialog if error occurs during update
- Show dialog to shutdown or reboot if needed

Peek 2021-01-15 20-20

@cassidyjames could you please have a look if you are asked to reboot your system.

It's getting difficult to test all cases without supported hardware 🙈

src/Widgets/FirmwareUpdateWidget.vala
```diff
@@ -0,0 +1,176 @@
...
+    private void show_details_dialog (Details details) {
+        var message_dialog = new Granite.MessageDialog.with_image_from_icon_name (
+            _("Manual steps required"),
+            details.caption,
+            "application-x-firmware",
+            Gtk.ButtonsType.NONE
+        );
+        message_dialog.transient_for = (Gtk.Window) get_toplevel ();
+
+        var suggested_button = new Gtk.Button.with_label (_("OK"));
```
danrabbit, line 118: We shouldn't use the label "OK" since it doesn't communicate a specific action. It's unclear from the button label what this will actually do.

src/Widgets/FirmwareUpdateWidget.vala
```diff
@@ -0,0 +1,176 @@
...
+        var message_dialog = new Granite.MessageDialog.with_image_from_icon_name (
+            _("An update requires the system to shutdown to complete"),
+            _("Shutdown now?"),
+            "application-x-firmware",
+            Gtk.ButtonsType.CANCEL
+        );
+        message_dialog.transient_for = (Gtk.Window) get_toplevel ();
+
+        var suggested_button = new Gtk.Button.with_label (_("Shutdown"));
+        suggested_button.get_style_context ().add_class (Gtk.STYLE_CLASS_SUGGESTED_ACTION);
```
danrabbit, line 165: This should probably be destructive action since it will cause all programs to close. Unless this will call the system shutdown dialog, then it's probably fine

```suggestion
        suggested_button.get_style_context ().add_class (Gtk.STYLE_CLASS_DESTRUCTIVE_ACTION);
```

src/Widgets/FirmwareUpdateWidget.vala
```diff
@@ -0,0 +1,176 @@
...
+            LoginManager.get_instance ().reboot ();
+        }
+
+        message_dialog.destroy ();
+    }
+
+    private void show_shutdown_dialog () {
+        var message_dialog = new Granite.MessageDialog.with_image_from_icon_name (
+            _("An update requires the system to shutdown to complete"),
+            _("Shutdown now?"),
```
danrabbit, line 158: Probably should use the same copy from the shutdown dialog in indicator session:
```suggestion
            _("This will close all open applications and turn off this device."),
```

src/Widgets/FirmwareUpdateWidget.vala
```diff
@@ -0,0 +1,176 @@
...
+*
+* You should have received a copy of the GNU General Public
+* License along with this program; if not, write to the
+* Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+* Boston, MA 02110-1301 USA
+*
+* Authored by: Marius Meisenzahl <mariusmeisenzahl@gmail.com>
+*/
+
+public class About.Widgets.FirmwareUpdateWidget : Gtk.ListBoxRow {
```
danrabbit, line 22: This should probably be more specific so that it's clear from the file name that this is the listboxrow
```suggestion
public class About.Widgets.FirmwareUpdateRow : Gtk.ListBoxRow {
```

src/Views/FirmwareView.vala
```diff
@@ -0,0 +1,133 @@
...
+        List<Device> devices = new List<Device> ();
+        foreach (var device in yield FwupdManager.get_instance ().get_devices ()) {
+            if (device.is (DeviceFlag.UPDATABLE) && device.releases.length () > 0) {
+                devices.append (device);
+            }
+        }
+
+        stack.visible_child = grid;
+
+        foreach (var device in devices) {
```
danrabbit: Why are we iterating over this list twice?

danrabbit (changes requested):
Just a few things that stick out doing a first pass

src/Views/FirmwareView.vala
meisenzahl: Good catch. Fixed this

src/Objects/DeviceFlag.vala
```diff
@@ -0,0 +1,363 @@
...
+                return null;
+            case INTERNAL:
+                /* TRANSLATORS: Device cannot be removed easily*/
+                return _("Internal device");
+            case UPDATABLE:
+                /* TRANSLATORS: Device is updatable in this or any other mode */
+                return _("Updatable");
+            case ONLY_OFFLINE:
+                /* TRANSLATORS: Update can only be done from offline mode */
+                return _("Update requires a reboot");
```
cassidyjames, line 221: We use "restart" instead of "reboot," and I want to clarify that it requires restarting the whole system, not just the device being updated. Typically we'd say "this device" to refer to the system, but that seems ambiguous in this context.

```suggestion
                return _("Update requires restarting the system");
```

src/Objects/DeviceFlag.vala
```diff
@@ -0,0 +1,363 @@
...
+                return _("Device is locked");
+            case SUPPORTED:
+                /* TRANSLATORS: Is found in current metadata */
+                return _("Supported on LVFS");
+            case NEEDS_BOOTLOADER:
+                /* TRANSLATORS: Requires a bootloader mode to be manually enabled by the user */
+                return _("Requires a bootloader");
+            case NEEDS_REBOOT:
+                /* TRANSLATORS: Requires a reboot to apply firmware or to reload hardware */
+                return _("Needs a reboot after installation");
```
cassidyjames, line 236: Same as above; is there any reason this should be translated differently? The end result is that it will require a restart.

```suggestion
                return _("Update requires restarting the system");
```

src/Objects/DeviceFlag.vala
```diff
@@ -0,0 +1,363 @@
...
+            case NEEDS_SHUTDOWN:
+                /* TRANSLATORS: Requires system shutdown to apply firmware */
+                return _("Needs shutdown after installation");
```
cassidyjames, line 239:
```suggestion
                return _("Requires the system to shut down after installation");
```

src/Views/FirmwareView.vala
```diff
@@ -0,0 +1,117 @@
...
+    private Gtk.Grid grid;
+    private Gtk.Grid progress_view;
+    private Gtk.ListBox update_list;
+
+    construct {
+        transition_type = Gtk.StackTransitionType.SLIDE_LEFT_RIGHT;
+
+        var progress_alert_view = new Granite.Widgets.AlertView (
+            _("Device is being updated"),
+            _("Do not unplug the device during the update."),
```
cassidyjames, line 32: Can we use the device name here so it's more clear? E.g.

>**“System Firmware” is being updated**
>Do not unplug the device during the update.

src/Widgets/FirmwareUpdateRow.vala
```diff
@@ -0,0 +1,176 @@
...
+
+        message_dialog.badge_icon = new ThemedIcon ("dialog-information");
+        message_dialog.show_all ();
+        message_dialog.run ();
+        message_dialog.destroy ();
+    }
+
+    private void show_reboot_dialog () {
+        var message_dialog = new Granite.MessageDialog.with_image_from_icon_name (
+            _("An update requires a reboot to complete"),
```
cassidyjames, line 135:
```suggestion
            _("An update requires the system to restart to complete"),
```

src/Widgets/FirmwareUpdateRow.vala
```diff
@@ -0,0 +1,176 @@
...
+        if (message_dialog.run () == Gtk.ResponseType.OK) {
+            LoginManager.get_instance ().reboot ();
+        }
+
+        message_dialog.destroy ();
+    }
+
+    private void show_shutdown_dialog () {
+        var message_dialog = new Granite.MessageDialog.with_image_from_icon_name (
+            _("An update requires the system to shutdown to complete"),
```
cassidyjames, line 157:
```suggestion
            _("An update requires the system to shut down to complete"),
```

src/Widgets/FirmwareUpdateRow.vala
```diff
@@ -0,0 +1,176 @@
...
+            _("This will close all open applications and turn off this device."),
+            "application-x-firmware",
+            Gtk.ButtonsType.CANCEL
+        );
+        message_dialog.transient_for = (Gtk.Window) get_toplevel ();
+
+        var suggested_button = new Gtk.Button.with_label (_("Shutdown"));
```
cassidyjames, line 164:
```suggestion
        var suggested_button = new Gtk.Button.with_label (_("Shut Down"));
```

src/Widgets/FirmwareUpdateRow.vala
```diff
@@ -0,0 +1,176 @@
...
+        suggested_button.get_style_context ().add_class (Gtk.STYLE_CLASS_DESTRUCTIVE_ACTION);
+        message_dialog.add_action_widget (suggested_button, Gtk.ResponseType.OK);
+
+        message_dialog.badge_icon = new ThemedIcon ("dialog-information");
```
cassidyjames, line 168:
```suggestion
        message_dialog.badge_icon = new ThemedIcon ("system-shutdown");
```

cassidyjames (changes requested):
A copy pass

src/Widgets/FirmwareUpdateRow.vala
meisenzahl: I have also changed it for reboot based on your suggestion

src/Widgets/FirmwareUpdateRow.vala
```diff
@@ -0,0 +1,176 @@
...
+    private void show_reboot_dialog () {
+        var message_dialog = new Granite.MessageDialog.with_image_from_icon_name (
+            _("An update requires the system to restart to complete"),
+            _("This will close all open applications and restart this device."),
+            "application-x-firmware",
+            Gtk.ButtonsType.CANCEL
+        );
+        message_dialog.transient_for = (Gtk.Window) get_toplevel ();
+
+        var suggested_button = new Gtk.Button.with_label (_("Reboot"));
```
cassidyjames, line 142: I missed a couple things

```suggestion
        var suggested_button = new Gtk.Button.with_label (_("Restart"));
```

cassidyjames, line 142: I missed a couple things

```suggestion
        var suggested_button = new Gtk.Button.with_label (_("Restart"));
```

cassidyjames (changes requested):
I was about to suggest another icon change, but it looks like you got it :)

danrabbit:
Something I noticed is that if I open the firmware page and then plug in a device, the view does not update. I have to close System Settings and reopen for it to show. If possible, it would be nice to automatically reload when new devices are plugged in. But if that's not possible we probably should have some kind of refresh button

src/Widgets/FirmwareUpdateRow.vala
```diff
@@ -0,0 +1,176 @@
...
+                show_reboot_dialog ();
+            } else if (device.is (DeviceFlag.NEEDS_SHUTDOWN)) {
+                show_shutdown_dialog ();
+            }
+        }
+    }
+
+    private void show_details_dialog (Details details) {
+        var message_dialog = new Granite.MessageDialog.with_image_from_icon_name (
+            _("Manual steps required"),
```
danrabbit, line 111: The CLI I think actually has a pretty good string here which is:
```suggestion
            _("“%s” needs to manually be put in update mode").printf (device.name),
```

src/Widgets/FirmwareUpdateRow.vala
```diff
@@ -0,0 +1,176 @@
...
+            details.caption,
+            "application-x-firmware",
```
danrabbit, line 113: This should probably be
```suggestion
            device.icon,
```

src/Views/FirmwareView.vala
```diff
@@ -0,0 +1,119 @@
...
+        visible_child = grid;
+
+        update_list.show_all ();
+    }
+
+    private void on_error (string error) {
+        var message_dialog = new Granite.MessageDialog.with_image_from_icon_name (
+            _("Failed to install firmware release"),
+            error,
+            "application-x-firmware",
```
danrabbit, line 110:
```suggestion
            device.icon,
```

danrabbit (changes requested):

meisenzahl:
## Current state

- Reload when new device is plugged in
- Reload when device is removed

Peek 2021-01-16 08-11

It sometimes takes a while until the insertion or removal of a device is recognized. However, this is due to the fact that the signals on the D-Bus are emitted late.

@danrabbit thanks for your attention to detail 🚀

src/Objects/Device.vala
```diff
@@ -0,0 +1,46 @@
...
+
+public class About.Device : Object {
+    public string id { get; set; }
+    public string name { get; set; }
+    public string summary { get; set; }
+    public string icon { get; set; }
+    public string vendor { get; set; }
+    public string version { get; set; }
+    public string[] guids { get; set; }
+    public uint64 flags { get; set; }
```
davidmhewitt, line 30: Should the type here be `DeviceFlag`?

src/Objects/Device.vala
```diff
@@ -0,0 +1,46 @@
...
+    public uint32 install_duration { get; set; }
+    public string update_error { get; set; }
+
+    public List<Release> releases { get; owned set; }
+    public Release latest_release { get { return releases.nth_data (0); }}
+
+    public bool is (DeviceFlag flag) {
```
davidmhewitt, line 37: Possibly better to call this method `has_flag`?

src/Objects/Device.vala
```diff
@@ -0,0 +1,46 @@
...
+        foreach (var f in DeviceFlag.get_list (flags)) {
```
davidmhewitt, line 38: This seems like a really odd way to see if a flag is set. Do we need the `get_list` method at all? This whole method is just equivalent to `return flag in flags` right?

src/Objects/ReleaseFlag.vala
```diff
@@ -0,0 +1,62 @@
...
+    TRUSTED_PAYLOAD = (1u << 0),     /* Since: 1.2.6 */
+    TRUSTED_METADATA = (1u << 1),    /* Since: 1.2.6 */
+    IS_UPGRADE = (1u << 2),          /* Since: 1.2.6 */
+    IS_DOWNGRADE = (1u << 3),        /* Since: 1.2.6 */
+    BLOCKED_VERSION = (1u << 4),     /* Since: 1.2.6 */
+    BLOCKED_APPROVAL = (1u << 5),    /* Since: 1.2.6 */
+    IS_ALTERNATE_BRANCH = (1u << 6), /* Since: 1.5.0 */
+    UNKNOWN = uint64.MAX;            /* Since: 1.2.6 */
+
+    public static ReleaseFlag from_uint64 (uint64 flag) {
```
davidmhewitt, line 34: Is this necessary? Can you not just directly cast the `uint64` to a `ReleaseFlag`? This also doesn't handle the case if multiple flags are set in the `uint64`, although I'm not sure if that's possible in the fwupd spec.

src/Objects/DeviceFlag.vala
```diff
@@ -0,0 +1,363 @@
...
+* You should have received a copy of the GNU General Public
+* License along with this program; if not, write to the
+* Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+* Boston, MA 02110-1301 USA
+*
+* Authored by: Marius Meisenzahl <mariusmeisenzahl@gmail.com>
+*/
+
+// https://github.com/fwupd/fwupd/blob/72df1147933de747312aa7c9892f07e7916b8a39/libfwupd/fwupd-enums.h#L133
+public enum About.DeviceFlag {
```
davidmhewitt, line 23: Add a `[Flags]` annotation above this line.

src/Objects/ReleaseFlag.vala
```diff
@@ -0,0 +1,62 @@
...
+* You should have received a copy of the GNU General Public
+* License along with this program; if not, write to the
+* Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+* Boston, MA 02110-1301 USA
+*
+* Authored by: Marius Meisenzahl <mariusmeisenzahl@gmail.com>
+*/
+
+// https://github.com/fwupd/fwupd/blob/72df1147933de747312aa7c9892f07e7916b8a39/libfwupd/fwupd-enums.h#L192
+public enum About.ReleaseFlag {
```
davidmhewitt, line 23: Add a `[Flags]` annotation above this line.

davidmhewitt (changes requested):
This is not a complete code review but I just had a quick glance and noticed some odd things about the way you are handling flag types. Let me know if there's a reason for this, otherwise I think it can be vastly simplified.

meisenzahl:
@davidmhewitt thank you very much for reviewing the code 🙏

I was not aware of Vala's `[Flags]` feature 🤯

davidmhewitt (changes requested):
Awaiting the rest of the DBus methods to be refactored to use a DBusProxy instead of the manual DBus connection. Will conduct another code review after that.

meisenzahl:
@davidmhewitt thanks for your review!

Code is now much cleaner.

I was not aware that I can annotate the D-Bus methods with `async` 🙈

src/Interfaces/FwupdManager.vala
```diff
@@ -0,0 +1,317 @@
...
+        if (device.id.length > 0 && device.has_flag (DeviceFlag.UPDATABLE)) {
+            device.releases = yield get_releases (device.id);
+        } else {
+            device.releases = new List<Release> ();
+        }
+
+        return device;
+    }
+
+    private async Release parse_release (GLib.HashTable<string, Variant> serialized_release) {
```
davidmhewitt, line 138: There aren't any async calls in this method, so you can remove `async`

src/fd.c
```diff
@@ -0,0 +1,5 @@
+#include <fcntl.h>
+
+int ro_fd (const char *path) {
+    return open (path, O_RDONLY);
```
davidmhewitt, line 4: `Posix.open` is the vala binding of the same method, any reason why you're exposing it from C?

src/Objects/Details.vala
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program; if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301 USA
+ *
+ * Authored by: Marius Meisenzahl <mariusmeisenzahl@gmail.com>
+ */
+
+public class About.Details : Object {
```
davidmhewitt, line 22: I think I'd prefer to see all of these class names prefixed with `Fwupd` so we know they're `Fwupd` types. Otherwise they're a bit too generic.

src/Interfaces/FwupdManager.vala
```diff
@@ -0,0 +1,317 @@
...
+            options.insert ("allow-older", new Variant.boolean (true));
+            options.insert ("allow-reinstall", new Variant.boolean (true));
+            options.insert ("no-history", new Variant.boolean (true));
+
+            var fd = ro_fd (path);
+            var handle = new UnixInputStream (fd, true);
+
+            yield fwupd.install (device.id, handle, options);
+        } catch (Error e) {
+            warning ("Could not connect to fwupd interface: %s", e.message);
```
davidmhewitt, line 254: This warning message is too generic, we're already connected to the fwupd interface, this is a failure to install a firmware update, right?

src/Interfaces/FwupdManager.vala
```diff
@@ -0,0 +1,317 @@
...
+    private async List<Release> get_releases (string id) {
+        var releases_list = new List<Release> ();
+
+        try {
+            var result = yield fwupd.get_releases (id);
+            foreach (unowned GLib.HashTable<string, Variant> release in result) {
+                releases_list.append (yield parse_release (release));
+            }
+        } catch (Error e) {
+            warning ("Could not connect to fwupd interface: %s", e.message);
```
davidmhewitt, line 77: Warning is too generic again

src/Interfaces/FwupdManager.vala
```diff
@@ -0,0 +1,317 @@
...
+    public async List<Device> get_devices () {
+        var devices_list = new List<Device> ();
+
+        try {
+            var result = yield fwupd.get_devices ();
+            foreach (unowned GLib.HashTable<string, Variant> device in result) {
+                devices_list.append (yield parse_device (device));
+            }
+        } catch (Error e) {
+            warning ("Could not connect to fwupd interface: %s", e.message);
```
davidmhewitt, line 62: Warning too generic

src/Interfaces/FwupdManager.vala
```diff
@@ -0,0 +1,317 @@
...
+* Boston, MA 02110-1301 USA
+*
+* Authored by: Marius Meisenzahl <mariusmeisenzahl@gmail.com>
+*/
+
+extern int ro_fd (string path);
+
+public class About.FwupdManager : Object {
+    [DBus (name = "org.freedesktop.fwupd")]
+    public interface FwupdInterface : Object {
```
davidmhewitt, line 26: This interface can probably be private, I don't think we use it outside of this class

src/Interfaces/FwupdManager.vala
```diff
@@ -0,0 +1,317 @@
...
+        public abstract async GLib.HashTable<string, Variant>[] get_details (UnixInputStream handle) throws GLib.Error;
+    }
+
+    private FwupdInterface fwupd;
+
+    public signal void on_device_added (Device device);
+    public signal void on_device_error (Device device, string error);
+    public signal void on_device_removed (Device device);
+
+    static FwupdManager? instance = null;
```
davidmhewitt, line 42: I think I'd prefer to not use a singleton for this class. It would be better to construct one instance of this class in `FirmwareView` and then use signals from `FirmwareUpdateRow` to trigger the update process back in `FirmwareView`. Or, if that turns out to be too complex, you could pass the reference from `FirmwareView` to `FirmwareUpdateRow`

src/Interfaces/FwupdManager.vala
```diff
@@ -0,0 +1,317 @@
...
+        return Path.build_path (Path.DIR_SEPARATOR_S, Environment.get_tmp_dir (), file_path);
+    }
+
+    public async string? download_file (Device device, string uri) {
+        var path = get_path (uri);
+
+        File server_file = File.new_for_uri (uri);
+        File local_file = File.new_for_path (path);
+
+        if (FileUtils.test (path, FileTest.IS_REGULAR)) {
```
davidmhewitt, line 219: I'm not sure I like this. What if we're downloading a file with a really common name like `image.jpg` and something else has already created that file in the tmp directory? We would return completely the wrong file. It's better to just remove this statement and let the `FileCopyFlags.OVERWRITE` deal with it.

src/Interfaces/FwupdManager.vala
```diff
@@ -0,0 +1,317 @@
...
+        try {
+            var fd = ro_fd (path);
+            var handle = new UnixInputStream (fd, true);
+
+            var result = yield fwupd.get_details (handle);
+            foreach (unowned GLib.HashTable<string, Variant> serialized_details in result) {
+                serialized_details.@foreach ((key, val) => {
+                    if (key == "Release") {
+                        var iter = val.iterator ().next_value ().iterator ();
+                        while (iter.next ("{sv}", out key, out val)) {
```
davidmhewitt, line 274: I don't like that we're overwrite `key` and `val` of the outer loop in an inner loop here. It's probably fine, but the code would be clearer if you used different variables.

src/Objects/DeviceFlag.vala
```diff
@@ -0,0 +1,225 @@
...
+    MD_SET_ICON = (1u << 41),            /* Since: 1.5.2 */
+    UNKNOWN = uint64.MAX;                /* Since: 0.7.3 */
+
+    // https://gitlab.gnome.org/hughsie/gnome-firmware-updater/-/blob/f5281078e3cfade7ff919c812ba63de22431aaf2/src/gfu-common.c#L249
+    public string? to_string () {
+        switch (this) {
+            case NONE:
+                return null;
+            case INTERNAL:
+                /* TRANSLATORS: Device cannot be removed easily*/
```
davidmhewitt, line 76: I think `TRANSLATORS` comments need to be line comments starting with `///` so that they're automatically added to the `.pot` files, so that will need doing throughout.

src/Widgets/FirmwareUpdateRow.vala
```diff
@@ -0,0 +1,176 @@
...
+
+    private void add_up_to_date_label (Gtk.Grid grid) {
+        var update_to_date_label = new Gtk.Label (_("Up to date")) {
+            valign = Gtk.Align.CENTER
+        };
+        grid.attach (update_to_date_label, 2, 0, 1, 2);
+    }
+
+    private async void update (Device device, Release release) {
+        var path = yield FwupdManager.get_instance ().download_file (device, release.uri);
```
davidmhewitt, line 92: I wonder if it's cleaner to add this logic to the `Release` class? So instead of needing the `FwupdManager` class here, you can just do `release.download ()`

src/Widgets/FirmwareUpdateRow.vala
```diff
@@ -0,0 +1,176 @@
...
+
+        var details = yield FwupdManager.get_instance ().get_details (device, path);
```
davidmhewitt, line 94: Again, is there some way you can switch this around so you can just do, `device.get_details (path)`. Are these details about the update or release? If so, we should probably rename the method to reflect that (i.e. `get_update_details` or `get_release_details`).

src/Widgets/FirmwareUpdateRow.vala
```diff
@@ -0,0 +1,176 @@
...
+
+        if (details.caption != null) {
+            show_details_dialog (details);
```
davidmhewitt, line 97: Would there be some expectation that we'd be able to cancel the upgrade in this dialog? We might not like some of the "details"? It looks like no matter what we do to this dialog, the upgrade happens anyway. Maybe that's an @elementary/ux question.

davidmhewitt (changes requested):
Another bunch of review comments!

src/Widgets/FirmwareUpdateRow.vala
danrabbit: Yeah that's a great point, there probably should be an option to cancel here

src/Interfaces/FwupdManager.vala
davidmhewitt: Or, if you want to efficiently check it's the same file and not download it again, you could check the timestamps like what AppCenter does in the screenshot cache.

src/Interfaces/FwupdManager.vala
davidmhewitt: Or if you efficiently want to check that it's the same file and not download it again, you could check the timestamps like what the screenshot cache in AppCenter does.

meisenzahl: I will address this on a follow up ticket. Fwupd provides checksums for the files.

src/Widgets/FirmwareUpdateRow.vala
meisenzahl: These details are release details. So I renamed it to `get_release_details`.

This method calls the D-Bus interface. So I would rather not move it to the `Device` class.

src/Widgets/FirmwareUpdateRow.vala
meisenzahl: This method emits the signal `on_device_error`. This signal passes information of the device.

This info is used to generate a user facing dialog. I could move the method to the `Device` class.

@davidmhewitt what is your opinion on this? Also in connection with your suggestion to refactor `get_details (path)`.

src/Interfaces/FwupdManager.vala
```diff
@@ -0,0 +1,302 @@
...
+            }
+        });
+
+        return release;
+    }
+
+    private string get_path (string uri) {
+        var parts = uri.split ("/");
+        string file_path = parts[parts.length - 1];
+        return Path.build_path (Path.DIR_SEPARATOR_S, Environment.get_tmp_dir (), file_path);
```
davidmhewitt, line 197: You can use `Path.build_filename` and then you don't need to include `Path.DIR_SEPARATOR_S`

src/Interfaces/FwupdManager.vala
```diff
@@ -0,0 +1,302 @@
...
+                    break;
+            }
+        });
+
+        return release;
+    }
+
+    private string get_path (string uri) {
+        var parts = uri.split ("/");
+        string file_path = parts[parts.length - 1];
```
davidmhewitt, line 196: `File.get_basename` is equivalent to this, so you should be able to use the `File` object you create in `download_file` and call `get_basename` on it.

src/Interfaces/FwupdManager.vala
```diff
@@ -0,0 +1,302 @@
...
+
+    public async string? download_file (Fwupd.Device device, string uri) {
+        var path = get_path (uri);
+
+        File server_file = File.new_for_uri (uri);
+        File local_file = File.new_for_path (path);
+
+        bool result;
+        try {
+            result = yield server_file.copy_async (local_file, FileCopyFlags.OVERWRITE, Priority.DEFAULT, null, (current_num_bytes, total_num_bytes) => {});
```
davidmhewitt, line 208: I think I'd like to see something done with this progress callback in the future so that people with slow connections or that are downloading large files know something is happening. I guess add a TODO comment for now and we'll open issues for all of them once this is merged.

src/Interfaces/FwupdManager.vala
```diff
@@ -0,0 +1,302 @@
...
+        try {
+            // https://github.com/fwupd/fwupd/blob/c0d4c09a02a40167e9de57f82c0033bb92e24167/libfwupd/fwupd-client.c#L2045
+            var options = new GLib.HashTable<string, Variant> (str_hash, str_equal);
+            options.insert ("reason", new Variant.string ("user-action"));
+            options.insert ("filename", new Variant.string (path));
+            options.insert ("allow-older", new Variant.boolean (true));
+            options.insert ("allow-reinstall", new Variant.boolean (true));
+            options.insert ("no-history", new Variant.boolean (true));
+
+            var fd = Posix.open (path, Posix.O_RDONLY);
```
davidmhewitt, line 232: Are you supposed to close this file descriptor once done, or does fwupd handle that?

src/Widgets/FirmwareUpdateRow.vala
```diff
@@ -0,0 +1,186 @@
...
+    private FwupdManager fwupd;
+    public Fwupd.Device device { get; construct set; }
+
+    public signal void on_update_start ();
+    public signal void on_update_end ();
+
+    public FirmwareUpdateRow (FwupdManager fwupd, Fwupd.Device device) {
+        Object (device: device);
+
+        this.fwupd = fwupd;
```
davidmhewitt, line 32: Can this be a property too? Like `device`

davidmhewitt (changes requested):
Last few comments

src/Widgets/FirmwareUpdateRow.vala
davidmhewitt: Yes, I see now. I was thinking about how we could try and make the Fwupd interface feel a bit more object orientated, but since the DBus interface hasn't really been written in that way, it just makes things more complicated. It looks slightly cleaner now we're not using a singleton anyway, so we can forget this for now!

src/Widgets/FirmwareUpdateRow.vala
davidmhewitt: Again, let's forget this for now, it just adds more complication for not much benefit.

src/Interfaces/FwupdManager.vala
meisenzahl: To test this I called `Posix.close` twice. The first time it returned `0`. The second time it returned `-1`. So I suppose we have to take care of closing it ourselves.

I refactored the code and added `Posix.close (fd);` in a `finally` statement.

danrabbit:
I'm getting a real big lag when first opening the System plug again :(

davidmhewitt:
@danrabbit ah, oops, I wonder if that's related to my merge earlier. Does it go away if you install `switcheroo-control`. It might be timing out waiting for that.

danrabbit:
@davidmhewitt ah yup, I think that's it. Created https://github.com/elementary/switchboard-plug-about/pull/171

davidmhewitt (approved):
I'm happy with the code, but I have not conducted any testing. Nice work @meisenzahl !

meisenzahl:
@davidmhewitt thank you for your reviews and constructive feedback.

danrabbit (approved):
From the testing I've done, I'm happy with this as an MVP. I think we can wait until follow ups to iterate on the design and add features. But this does the thing as far as I'm concerned:)

meisenzahl:
I created some issues:

- [Add detail page for firmware releases](https://github.com/elementary/switchboard-plug-about/issues/172)
- [Show useful information during firmware update](https://github.com/elementary/switchboard-plug-about/issues/173)
- [Add cache for firmware updates](https://github.com/elementary/switchboard-plug-about/issues/174)

Feel free to create some more if I missed something :)

cassidyjames (approved):
:shipit:

PR #153, opened 2021-01-04
Split Hardware and OS views into tabs

danrabbit:
So that #150 is purely about adding a firmware view and also doesn't include this refactor

### DIFF
--- a/po/POTFILES
+++ b/po/POTFILES
@@ -1,2 +1,3 @@
 src/Plug.vala
 src/Views/HardwareView.vala
+src/Views/OperatingSystemView.vala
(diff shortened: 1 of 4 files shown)


cassidyjames:
@danrabbit alternatively we could just have an "About" view with the current plug contents, and then a new "Firmware" view. I'm not sure splitting these up is needed without a larger redesign… but perhaps we do this step and then think about how to make them look as nice as possible when we get the whole page to work with?

danrabbit:
@cassidyjames Yeah, something I'd like to do in a follow up branch is to move the buttons on the OS view to a vertical stack. I've seen in some localizations that these buttons are super cramped and can even force the window to expand

cassidyjames:
@danrabbit I'm poking at a larger layout reworking in a branch of this (which can be rebased on master if this is merged). But this looks like a good first step.

cassidyjames (approved):

PR #154, opened 2021-01-04
OperatingSystemView: Redesign to be laid out side-by-side

cassidyjames:
I figure if we like this sort of layout, I can do the hardware view next in a follow-up. Needs https://github.com/elementary/stylesheet/issues/921 to look perfect, but shouldn't really be a blocker.

Screenshot from 2021-01-04 15-04-34


danrabbit (approved):
:rocket:

PR #158, opened 2021-01-05
OSView: Overlay logo on default wallpaper

danrabbit:
Screenshot from 2021-01-05 10 37 21@2x

### DIFF
--- a/data/OperatingSystemView.css
+++ b/data/OperatingSystemView.css
@@ -1,5 +1,9 @@
 avatar.image {
-    background-color: alpha(black, 0.05);
+    background-image: linear-gradient(
+        to bottom,
+        alpha(@accent_color_500, 0.25),
+        alpha(@accent_color_700, 0.75)
+    );
 }
 
 .logo {
--- a/data/OperatingSystemView.css
+++ b/data/OperatingSystemView.css
@@ -9,6 +9,6 @@ avatar.image {
 .logo {
     color: white;
     -gtk-icon-shadow:
-        0 1px 1px alpha(black, 0.2),
-        0 2px 3px alpha(black, 0.15);
+        0 1px 1px alpha(black, 0.3),
+        0 2px 3px alpha(@accent_color_900, 0.2);
 }
--- a/data/OperatingSystemView.css
+++ b/data/OperatingSystemView.css
@@ -1,0 +1,10 @@
+avatar.image {
+    background-color: alpha(black, 0.05);
+}
+
+.logo {
+    color: white;
+    -gtk-icon-shadow:
+        0 1px 1px alpha(black, 0.2),
+        0 2px 3px alpha(black, 0.15);
+}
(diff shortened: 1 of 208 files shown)


src/Views/OperatingSystemView.vala
```diff
@@ -127,13 +152,13 @@ public class About.OperatingSystemView : Gtk.Grid {
         button_grid.set_child_secondary (settings_restore_button, true);
 
         var software_grid = new Gtk.Grid () {
-            column_spacing = 12,
+            column_spacing = 6,
```
cassidyjames, line 155: I was actually wondering if we should go in the opposite direction; as-is, this large graphic is all up on the text; I was playing with 24 locally and that felt like it let the logo breathe a little better.

```suggestion
            column_spacing = 24,
```

Screenshot from 2021-01-04 19-05-12

src/Views/OperatingSystemView.vala
```diff
@@ -53,11 +56,33 @@ public class About.OperatingSystemView : Gtk.Grid {
...
+                return new Gdk.Pixbuf.from_file_at_scale ("/usr/share/backgrounds/elementaryos-default", -1, size, true);
+            } catch (Error e) {
+                critical (e.message);
+            }
+        });
+        logo.get_style_context ().add_provider (style_provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
+
+        var icon = new Gtk.Image () {
+            icon_name = logo_icon_name + "-symbolic",
+            pixel_size = 122
```
cassidyjames, line 76: Might be worth commenting in case someone later wonders where the magic number came from, e.g.

```suggestion
            // 128 minus 3px padding on each side
            pixel_size = 122
```

data/OperatingSystemView.css
```diff
@@ -0,0 +1,10 @@
+avatar.image {
+    background-color: alpha(black, 0.05);
```
cassidyjames, line 2: I think a little more foreground protection would be best here, maybe

```suggestion
    background-color: alpha(black, 0.15);
```

cassidyjames (changes requested):
Looks mostly good to me, with a few comments here.

data/OperatingSystemView.css
cassidyjames: we also discussed trying this in Slack

```suggestion
    background-image: linear-gradient(
        to bottom,
        alpha(@accent_color_500, 0.25),
        alpha(@accent_color_700, 0.75)
    );
```

src/Views/OperatingSystemView.vala
```diff
@@ -129,13 +155,14 @@ public class About.OperatingSystemView : Gtk.Grid {
         button_grid.set_child_secondary (settings_restore_button, true);
 
         var software_grid = new Gtk.Grid () {
-            column_spacing = 12,
+            // The avatar has some built-in margin for shadows
+            column_spacing = 48 - 6,
```
cassidyjames, line 159: You'd mentioned wanting to drop this even more to 32px because of the illusion with the round image. I do think it would look better. 

```suggestion
            column_spacing = 48 - 6,
```

cassidyjames (approved):
Looks good, plus one comment you'd mentioned in Slack

PR #159, opened 2021-01-05
OSView: Add last update info when available

danrabbit:
Screenshot from 2021-01-04 17 12 03@2x

### DIFF
--- a/src/Views/OperatingSystemView.vala
+++ b/src/Views/OperatingSystemView.vala
@@ -148,7 +148,6 @@ public class About.OperatingSystemView : Gtk.Grid {
             software_grid.attach (last_update_label, 1, 1, 3);
         }
 
-
         if (upstream_release != null) {
             var based_off = new Gtk.Label (_("Built on %s").printf (upstream_release)) {
                 selectable = true,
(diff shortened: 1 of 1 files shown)


src/Views/OperatingSystemView.vala
```diff
@@ -104,12 +104,20 @@ public class About.OperatingSystemView : Gtk.Grid {
...
+        Gtk.Label? last_update_label = null;
         var appcenter_info = new GLib.DesktopAppInfo ("io.elementary.appcenter.desktop");
         if (appcenter_info != null) {
             update_button = new Gtk.Button.with_label (_("Check for Updates"));
             update_button.clicked.connect (() => {
                 appcenter_info.launch_action ("ShowUpdates", new GLib.AppLaunchContext ());
             });
+
+            var appcenter_settings = new GLib.Settings ("io.elementary.appcenter.settings");
+            var update_time = new GLib.DateTime.from_unix_utc (appcenter_settings.get_int64 ("last-refresh-time"));
```
vjr, line 116: How would this behave if there was never an update done via AppCenter? Say, perhaps on a new installation?

cassidyjames (changes requested):
I mentioned in Slack that this is the last time AppCenter _checked_ for an update, not necessarily the last time any updates were run. Since AppCenter does that daily, this is not likely to provide any useful information.

danrabbit:
Gonna close this and open an issue report instead

PR #167, opened 2021-01-15
Danrabbit/fwupd placeholder

danrabbit:
Use the built-in placeholder functionality of the listbox for the empty state

Uses title case and removes the duplicate icon

Screenshot from 2021-01-14 17 56 26@2x


meisenzahl (approved):
@danrabbit if I get rid of the header in https://github.com/elementary/switchboard-plug-about/pull/150, should I add an icon again in the placeholder?

Otherwise this looks good to me 👍

cassidyjames:
@meisenzahl yeah, we don't need the header with the tabbed layout, and in that case it would look nice in the placeholder.

PR #168, opened 2021-01-19
FwupdManager: Refactor get_devices

davidmhewitt:
This is a PR into an already open PR...

@meisenzahl I've opened this as a PR because it would have been quite complicated to explain in a code review :sweat_smile: 

Can you test this still works (I don't have any devices that support fwupd) ?

If so, you should be able to do the same for pretty much all your DBus methods and the code will be a lot cleaner. Ping me on Slack if you need any help!


meisenzahl (approved):
This works as intended and makes the code much cleaner.

PR #169, opened 2021-01-21
Use switcheroo-control to get dual GPU information

davidmhewitt:
Supersedes #141 

Also offers some improvements in the information we can get from certain GPUs. Haven't tested on a wide variety of hardware yet.

Before:
Screenshot from 2021-01-21 17-38-43

After:
Screenshot from 2021-01-21 17-40-32

Co-authored-by: Justin Haygood <justin.haygood@hudsonmx.com>


src/Views/HardwareView.vala
```diff
@@ -233,6 +255,36 @@ public class About.HardwareView : Gtk.Grid {
...
+        // and return null
+        if (!primary) {
+            return null;
+        }
+
+        if (session_manager != null) {
+            return clean_name (session_manager.renderer);
+        }
+
+        return _("Unknown GPU");
```
tintou, line 285: Do we want something more meaningful like `_("Primary GPU")` or `_("Secondary GPU")` so that it doesn't look broken :thinking:

davidmhewitt: I've just noticed I've changed the string from "Unknown Graphics" to "Unknown GPU" by accident. So I've changed that back.

But it's super unlikely that we'd hit this after querying two different sources for GPU information. I still think I'd rather see:

```
Quad-Core Intel Core i7-XXXX
Unknown Graphics
16 GB memory
128 GB storage
```

Instead of 
```
Quad-Core Intel Core i7-XXXX
Primary GPU/Graphics
16 GB memory
128 GB storage
```

tintou (approved):

PR #180, opened 2021-01-26
Update POTFILES and correct gettext syntax

ryonakano:
So that we can show the new firmware view in translated strings correctly


davidmhewitt (approved):
Thank you!

Issue #172, opened 2021-01-24
Add detail page for firmware releases


meisenzahl:
## Problem

Currently I only see a list of available firmware updates. But I don't know what the respective firmware update changes in detail.

![Bildschirmfoto von 2021-01-24 08 25 42](https://user-images.githubusercontent.com/10796736/105623911-cb220900-5e1d-11eb-8746-90b5235bb949.png)

## Proposal

Add a detail page for firmware releases.

## Prior Art

This is a screenshot of [GNOME Firmware](https://gitlab.gnome.org/hughsie/gnome-firmware-updater) showing available information of a release.

![Bildschirmfoto von 2021-01-24 08 21 38](https://user-images.githubusercontent.com/10796736/105623847-4f27c100-5e1d-11eb-983c-bae9d25b3dec.png)

cassidyjames:
@meisenzahl as an initial pass, clicking the row of an update should switch a stack to this view. At a glance, we probably want to include:

- Version
- Summary
- Description
- Vendor
- Size
- Install duration

Rough whiteboard sketch:

![PXL_20201214_193021427 NIGHT](https://user-images.githubusercontent.com/611168/105778633-f72ab080-5f29-11eb-8837-a9e6fd08299c.jpg)


Rough ASCII mockup (inside the stack):

```
[← All Updates ]    Unifying Receiver    [ Install ]
----------------------------------------------------

 ICON  Unifying Receiver
 ICON  A miniaturised USB wireless receiver


Summary here, maybe in bold

Description goes here; it's usually pretty long-form
and includes information about what's in this
release of the update, so we should show it up top.

                  Version: RQR24.06_B0030
                   Vendor: Logitech
                     Size: 56.8 kB
Estimated time to install: 30 sec
```

Honestly we probably don't _need_ the dialog-style icon, primary text, and secondary text up top since we can have the name of the device in the bar up top (see the installer, initial-setup, or adding a keyboard layout in the Keyboard plug for the sort of pattern here).

PR #182, opened 2021-01-26
Add detail page for firmware releases

meisenzahl:
Closes https://github.com/elementary/switchboard-plug-about/issues/172


meisenzahl:
## Current state

Peek 2021-01-26 20-30

I am so bad at doing layouts 🙈

@cassidyjames I could use some help 🙏

cassidyjames (changes requested):
The actual contents of the detail page looks close, and I can probably do any small layout cleanup once we're closer. I have a couple of thoughts though:

Formatter seems interesting; I would think some if not all of these would be built into GLib or something and not require manual handling here. Dealing with bytes and seconds and XML all seem pretty common. :stuck_out_tongue: So that might be worth looking into instead of reinventing it here.

Layout-wise, as I mentioned in Slack, ideally the detail page would be in a Hdy.Deck and within the same container as the list (so styled with the white background and border and everything); this way it feels like a page within that list, instead of not really having a clear container. See the Installer and Initial Setup lists where we use Hdy.Deck for an example of that (the right half of this view):

Screenshot from 2021-01-26 13 39 03

Screenshot from 2021-01-26 13-39-12

meisenzahl:
@cassidyjames layout-wise like this?

Peek 2021-01-26 22-04

cassidyjames:
@meisenzahl yeah that's much closer! Looks like we might need a style class or something on that top bar (see: Installer/Initial Setup to match)?

meisenzahl:
Bildschirmfoto von 2021-01-26 22 15 07

Based on https://github.com/elementary/initial-setup/blob/master/src/Widgets/VariantWidget.vala

src/Views/FirmwareReleaseView.vala
```diff
@@ -0,0 +1,172 @@
...
+        vendor_value_label.label = release.vendor;
+        size_value_label.label = Formatter.bytes_to_string (release.size);
+        install_duration_value_label.label = Formatter.seconds_to_string (release.install_duration);
+
+        show_all ();
+    }
+
+    construct {
+        orientation = Gtk.Orientation.VERTICAL;
+        row_spacing = 8;
```
cassidyjames, line 71: We should use margins on the widgets we want to have margins instead of this spacing.

```suggestion
```

src/Views/FirmwareReleaseView.vala
```diff
@@ -0,0 +1,172 @@
...
+                });
+
+                break;
+            default:
+                update_button.label = _("Up to date");
+                update_button.sensitive = false;
+                break;
+        }
+
+        title_label.label = device.name;
```
cassidyjames, line 58:
```suggestion
        title_label.label = "<b>%s</b>".printf (device.name);```

File: src/Views/FirmwareReleaseView.vala [hunk 3]
```diff
@@ -0,0 +1,172 @@
...
+        get_style_context ().add_class (Gtk.STYLE_CLASS_VIEW);
+
+        var back_button = new Gtk.Button.with_label (_("All Updates")) {
+            halign = Gtk.Align.START,
+            margin = 6
+        };
+        back_button.get_style_context ().add_class (Granite.STYLE_CLASS_BACK_BUTTON);
+
+        title_label = new Gtk.Label ("") {
+            ellipsize = Pango.EllipsizeMode.END
```
cassidyjames, line 81:
```suggestion
            ellipsize = Pango.EllipsizeMode.END,
            use_markup = true
```

cassidyjames (changes requested):
Just a couple of style things to better match the other places we use this type of layout

meisenzahl:
> Formatter seems interesting; I would think some if not all of these would be built into GLib or something and not require manual handling here. Dealing with bytes and seconds and XML all seem pretty common. :stuck_out_tongue: So that might be worth looking into instead of reinventing it here.

@cassidyjames I refactored the code to use `GLib.format_size` to pretty print Bytes. But I could not find a function to pretty print time.

Regarding XML: I could add `libxml2-dev` as a dependency. Is there anything against it?

src/Views/FirmwareReleaseView.vala
```diff
@@ -0,0 +1,195 @@
...
+        description_label.label = release.description;
+        version_value_label.label = release.version;
+        vendor_value_label.label = release.vendor;
+        size_value_label.label = GLib.format_size (release.size);
+
+        if (release.install_duration < 60) {
+            install_duration_value_label.label = _("less than a minute");
+        } else {
+            install_duration_value_label.label = _("%llu minutes".printf (release.install_duration));
+        }
```
tintou, line 69:
```suggestion
        uint32 duration_minutes = release.install_duration / 60;
        if (duration_minutes < 1) {
            install_duration_value_label.label = _("less than a minute");
        } else {
            install_duration_value_label.label = GLib.ngettext ("%llu minute", "%llu minutes", duration_minutes).printf (duration_minutes);
        }
```
This fixes two issues regarding the translation:
 * `_("%d minutes")` doesn't work in all languages and requires to use `ngettext("singular", "plural", number)` as some languages have more than two cases when dealing with enumerating numbers.
 * `_("File %s could not being found".printf (file_name))` → `_("File %s could not being found").printf (file_name)` as translating a string works with a dictionary so you need to translate it first and replace the %s/%d after.

meisenzahl: @tintou thank you! Silly me 🙈

cassidyjames (dismissed):
UX-wise I'm happy with this!

src/Views/FirmwareView.vala
```diff
@@ -148,4 +172,116 @@ public class About.FirmwareView : Gtk.Stack {
...
+
+    private void show_release (Gtk.ListBoxRow widget) {
+        if (widget is Widgets.FirmwareUpdateRow) {
+            var row = (Widgets.FirmwareUpdateRow) widget;
+            firmware_release_view.update_view (row.device, row.device.latest_release);
+            deck.visible_child = firmware_release_view;
+
+            firmware_release_view.back.connect (() => {
+                deck.navigate (Hdy.NavigationDirection.BACK);
+            });
```
danrabbit, line 184: This can probably be connected in the construct block. It doesn't really make sense to connect this every time the view contents change since the back button will never change

danrabbit (changes requested):
I haven't read through all the code yet, but noticed this thing

meisenzahl:
## Current state

### TODO

- [ ] Add placeholder for devices without releases in `FirmwareReleaseView`
- [ ] Hide button in `header_box` if not upgradable or latest release is installed

Bildschirmfoto von 2021-02-09 06 54 07

meisenzahl:
https://user-images.githubusercontent.com/10796736/107959946-51a4b300-6fa4-11eb-964f-c7383d4e09d8.mp4

Should all be working now :grinning:

danrabbit:
I think I would prefer if we merged https://github.com/elementary/switchboard-plug-about/pull/202 first so that we're not making it harder to port to libfwupd

danrabbit:
We have fwupd docs now on Valadoc :) https://valadoc.org/fwupd/Fwupd.html

meisenzahl:
The way libfwupd's logic works, no release information is now displayed after a updatable device is updated to it's latest release...

Compare with my last video for Unifying Receiver.

src/Views/FirmwareView.vala
```diff
@@ -166,6 +194,33 @@ public class About.FirmwareView : Gtk.Stack {
...
+            deck.visible_child = firmware_release_view;
+
+            firmware_release_view.update.connect ((device, release) => {
+                on_update_start (device);
+
+                update.begin (device, release, (obj, res) => {
+                    update.end (res);
+                    on_update_end ();
+                });
+            });
```
danrabbit, line 210: This should be connected when the firmware release view is constructed, otherwise you're creating a new signal connection every time the release view is shown

src/Views/FirmwareView.vala
```diff
@@ -117,22 +135,32 @@ public class About.FirmwareView : Gtk.Stack {
...
+                    num_updates++;
+                }
+
+                update_list.add (row);
+                update_list.invalidate_sort ();
+                update_list.show_all ();
+
+                row.update.connect ((device, release) => {
+                    progress_alert_view.title = _("“%s” is being updated").printf (device.get_name ());
+                    visible_child = progress_view;
```
danrabbit, line 162: It looks like you created a function for this, but then didn't use it here?

danrabbit (changes requested):
Just stuff I noticed on a first pass here

src/Widgets/FirmwareUpdateRow.vala
```diff
@@ -20,19 +20,23 @@
...
-    public bool is_updatable { get; private set; default = false; }
+    public Fwupd.Release? release { get; construct set; }
 
-    public signal void on_update_start ();
-    public signal void on_update_end ();
+    public bool is_updatable {
+        get {
+            return release != null && device.get_version () != release.get_version ();
+        }
+    }
```
danrabbit, line 30: Why was this done? We already check this on construct, this change means we have to run this comparison every time this property is used

danrabbit (commented):
There's a lot going on here with so many things moving up into the FirmwareView, I wonder if it would make sense to split this into two PRs: one dealing with only the refactoring of the update logic and a separate one for adding the release view

meisenzahl:
@danrabbit sure let's break this up into separate PRs :+1:

PR #184, opened 2021-01-26
FirmwareReleaseView: Layout tweaks

cassidyjames:
Some layout tweaks, including adding a scrolled window to ensure things don't grow out of hand.

Screenshot from 2021-01-26 15-17-03


meisenzahl (approved):
LGTM

PR #186, opened 2021-01-27
FwupdManager: Use an async future to initialize dbus

davidmhewitt:
If fwupd isn't running (which you can test with `sudo killall fwupd`), there's a noticeable delay while loading the about plug while it waits synchronously for the DBus interface to be ready. This reduces that significantly.

@meisenzahl Can you check the `device_added` and `device_removed` signals still fire properly?


meisenzahl:
I can confirm that the signals still fire properly.

meisenzahl (approved):
I can confirm that this works as intended.

The implementation looks reasonable to me. But I lack the knowledge to judge it quality-wise.

PR #187, opened 2021-01-27
HardwareView: Get GPU info asynchronously

davidmhewitt:
Don't block the constructor (and hence the plug loading) by using synchronous DBus calls.

### DIFF
--- a/src/Views/HardwareView.vala
+++ b/src/Views/HardwareView.vala
@@ -22,8 +22,6 @@
 
 public class About.HardwareView : Gtk.Grid {
     private bool oem_enabled;
-    private string primary_gpu;
-    private string? secondary_gpu;
     private string hdd;
     private string manufacturer_icon_path;
     private string manufacturer_name;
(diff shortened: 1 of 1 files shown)


src/Views/HardwareView.vala
```diff
@@ -289,6 +298,18 @@ public class About.HardwareView : Gtk.Grid {
...
     }
 
+    private async void get_graphics_info () {
+        var primary_gpu = yield get_gpu_info (true);
+        primary_graphics_info.label = primary_gpu;
+
+        var secondary_gpu = yield get_gpu_info (false);
+        if (secondary_gpu != null) {
+            secondary_graphics_info.label = secondary_gpu;
+            graphics_grid.add (secondary_graphics_info);
```
danrabbit, line 308: Another way to do this without adding an additional grid would be to use https://valadoc.org/gtk+-3.0/Gtk.Grid.attach_next_to.html

danrabbit (changes requested):
I think we can save a grid with `attach_next_to` :)

src/Views/HardwareView.vala
davidmhewitt: Hm, I tried that, but it doesn't seem to update the items below it in the grid, so you just end up with two labels on top of each other.

davidmhewitt: @danrabbit Here's what I get. The second GPU label correctly attaches below the first, but doesn't push down the subsequent RAM and storage labels:

Screenshot from 2021-01-27 17-29-05

danrabbit: Oh you're right. That's lame. Nevermind then!

danrabbit (approved):

PR #189, opened 2021-01-27
HardwareView: Asyncify storage methods

davidmhewitt:
Get rid of all this blocking I/O in the constructor


danrabbit (approved):
Nice use of GLib.Subprocess as well

PR #191, opened 2021-01-28
Replace dialog.run with dialog.response.connect

danrabbit:
Can be rebase merged

`Gtk.Dialog.run ()` doesn't exist in Gtk4, so get ahead of that


src/Views/OperatingSystemView.vala
```diff
@@ -232,8 +232,13 @@ public class About.OperatingSystemView : Gtk.Grid {
...
-        dialog.destroy ();
+        dialog.show_all ();
+
+        int result = Gtk.ResponseType.NONE;
+        dialog.response.connect ((response_id) => {
+            result = response_id;
+            dialog.destroy ();
+        });
 
         return result == Gtk.ResponseType.ACCEPT;
```
tintou, line 243: This doesn't work as you're returning before the use responded

PR #194, opened 2021-02-05
Display memory size in units based on powers of 2

andreasfelix:
It might be a good idea to display the memory size using [prefixes based on the power of 2](https://en.wikipedia.org/wiki/Binary_prefix) instead of using SI prefixes, as these are the units most manufacturers use to specify RAM sizes.

This might also help to clear up some confusion: For example, my system has 16 GiB RAM and the hardware info displays 16.6 GB - which is technically correct - but not everyone may be aware that 1 GiB > 1 GB.

This would also be more in line with other programs, which also use the binary prefixes to display RAM sizes like gnome-control-center, htop, neofetch, etc.


cassidyjames:
Hm, on my laptop advertised with 16 GB of RAM, this branch shows `15.3 GiB memory` while master shows `16.4 GB memory`. I'm not sure which is _better_, per-say, but I still wonder if rounding to the nearest GB would match what most people expect.

andreasfelix:
To my (limited) understanding there are two reason the RAM size is not a round number:

1. `1 GiB = 1 <<  30 B` and `1 GB = 10^9 B` => `(1 << 30) / (10 ** 9) = 1.0737` => displaying a round number of `GiBs` in `GBs` will result in a 7% larger non-round number.
2. As far as I know, onboard-GPUs share the memory with the CPU => leads to a smaller number because some memory is used by the GPU.

This branch only "fixes" 1.

andreasfelix:
Even if you decide it is a good idea to round the number, you should round using the binary prefixes. Otherwise you will also get confusing results:

```
>>> for size in 4, 8, 16, 32, 64:
...    f"Real size: {size:2} GiB - Displayed: {ceil(size * 1.07)} GB"
... 
'Real size:  4 GiB - Displayed: 5 GB'
'Real size:  8 GiB - Displayed: 9 GB'
'Real size: 16 GiB - Displayed: 18 GB'
'Real size: 32 GiB - Displayed: 35 GB'
'Real size: 64 GiB - Displayed: 69 GB'

```

davidmhewitt:
Yes, it would be correct to use IEC units here, I think. 

RAM modules are sold and measured in GiB, so we should display using the same units. As @andreasfelix says, the current units we're using will look more wrong with higher amounts of RAM.

As for the discrepancy of amounts, I believe GLibTop is giving us the amount of memory reported by the kernel, which excludes any amounts reserved by the kernel itself or any GPUs or other hardware. Rounding may work as a solution to try and get closer to the actual installed hardware amount, but we'd have to be careful about how it was done to ensure we didn't make it worse.

It's worth noting that if we merge this, it'll be identical to how GNOME Control Center works and will display the exact same amount. So there's precedent for showing this amount of RAM on other linux distros, even if it can be slightly less than the hardware total sometimes.

davidmhewitt (approved):

Issue #193, opened 2021-02-04
Firmware: show devices without releases


cassidyjames:
## Problem

If I have supported devices but no releases for some reason (like unreachable fwupd servers), we don't currently show any devices. This can lead the user to believe no devices support firmware updates.

## Proposal

Instead, we should show all fwupd-enabled devices even if they don't have releases. We could show "No updates" instead of "Up to date" for the device if we want to be more precise.

## Prior Art

GNOME Firmware shows devices even without releases.

![Screenshot from 2021-02-04 13-51-22@2x](https://user-images.githubusercontent.com/611168/106953829-18d92400-66f0-11eb-9548-38ea9ec9da49.png)

cassidyjames:
@meisenzahl mentioned we could split this up with headings like we do in AppCenter with "Updates" and "Up to Date"

PR #195, opened 2021-02-05
Show devices without releases

meisenzahl:
Closes https://github.com/elementary/switchboard-plug-about/issues/193


meisenzahl:
Simplest solution without splitting items.

Bildschirmfoto von 2021-02-05 19 18 10

hanaral:
You might want to put those 'up to date' labels inside insensitive buttons, for visual consistency with the fact that the button UI elements are extrusions from the panel that can't just disappear. It also makes is consistent with other similar menus like Bluetooth or Network.

meisenzahl:
@hanaral thanks for your input. I already changed the behavior in https://github.com/elementary/switchboard-plug-about/pull/182

meisenzahl:
Split up with headings

Bildschirmfoto von 2021-02-05 19 42 45

Have to test some corner cases

cassidyjames:
@meisenzahl I think if we have headings, we don't need the individual labels here. It is redundant and would be cleaner without—plus it would make the action of updating firmware stand out even more.

cassidyjames:
@meisenzahl idk if you're still actively working on this, but as of the latest commit it looks like I've lost the "Up to Date" heading for my device that's up to date. :smile:

meisenzahl:
@cassidyjames I'm still tackling some issues as you might see 😛

Bildschirmfoto von 2021-02-05 21 04 20

src/Widgets/FirmwareHeaderRow.vala
```diff
@@ -0,0 +1,38 @@
...
+
+    public FirmwareHeaderRow.updatable (uint num_updates) {
+        Object (
+            label: ngettext ("%u Update Available", "%u Updates Available", num_updates).printf (num_updates)
+        );
+    }
+
+    public FirmwareHeaderRow.up_to_date () {
+        Object (label: _("Up to Date"));
+    }
```
danrabbit, line 37: This seems kind of overkill when it's pretty easy to do:
```
var updateable_header = new FirmwareHeaderRow (
    ngettext ("%u Update Available", "%u Updates Available", num_updates).printf (num_updates)
);
```
and
```
var updated_header = new FirmwareHeaderRow (_("Up to Date"));
```

Since this is such a simple class, it can probably just be private to the FirmwareView without these additional constructors

src/Views/FirmwareView.vala
```diff
@@ -143,9 +160,38 @@ public class About.FirmwareView : Gtk.Stack {
...
+                }
             }
         }
 
+        update_row_headers ();
+
         update_list.show_all ();
     }
+
+    private void update_row_headers () {
```
danrabbit, line 175: This should use https://valadoc.org/gtk+-3.0/Gtk.ListBox.set_header_func.html instead of trying to manually insert headers like this

src/Views/FirmwareView.vala
```diff
@@ -96,9 +105,17 @@ public class About.FirmwareView : Gtk.Stack {
...
-            update_list.add (row);
+
+            int position = -1;
+            if (device.releases.length () > 0 && device.latest_release.version != device.version) {
+                num_updatable_devices++;
+                position = 1;
+            }
+            update_row_headers ();
+
+            update_list.insert (row, position);
```
danrabbit, line 118: Instead of this, there should be a https://valadoc.org/gtk+-3.0/Gtk.ListBox.set_sort_func.html

danrabbit (changes requested):

meisenzahl:
@danrabbit thanks for your suggestions! The code is now much better structured.

However, I now have the problem that the header with the number of available updates is not updated. Do you have an idea how to implement this?

I have been following [elementary/switchboard-plug-bluetooth](https://github.com/elementary/switchboard-plug-bluetooth/blob/c6b5d8d825d50179faa04d51aea49ba69dcb22e5/src/MainView.vala) and [elementary/appcenter](https://github.com/elementary/appcenter/blob/e3a9b4feb4372621e53f96e56a6240c490548d43/src/Views/AppListUpdateView.vala). Bluetooth plug does not show a count of available devices and AppCenter always has the total count at the time the header is added.

meisenzahl:
I think I got it 🤔

https://user-images.githubusercontent.com/10796736/107112290-d120dc80-6856-11eb-838c-f5214052e999.mp4

danrabbit (approved):
Nice job!

PR #198, opened 2021-02-05
FirmwareView: Use native sort and header func

danrabbit:
Also make rows cache their own updatable state since they already check this when showing the update button

### DIFF
--- a/src/Views/FirmwareView.vala
+++ b/src/Views/FirmwareView.vala
@@ -60,6 +60,8 @@ public class About.FirmwareView : Gtk.Stack {
             selection_mode = Gtk.SelectionMode.SINGLE
         };
         update_list.set_placeholder (no_devices_alert_view);
+        update_list.set_sort_func ((Gtk.ListBoxSortFunc) sort_func);
+        update_list.set_header_func ((Gtk.ListBoxUpdateHeaderFunc) header_func);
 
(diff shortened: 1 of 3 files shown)


meisenzahl:
Just saw this after implementing it myself 🙈

meisenzahl:
Unfortunately, the same problem I encountered.

Bildschirmfoto von 2021-02-06 08 26 05

> @danrabbit thanks for your suggestions! The code is now much better structured.

> However, I now have the problem that the header with the number of available updates is not updated. Do you have an idea how to implement this?

> I have been following [elementary/switchboard-plug-bluetooth](https://github.com/elementary/switchboard-plug-bluetooth/blob/c6b5d8d825d50179faa04d51aea49ba69dcb22e5/src/MainView.vala) and [elementary/appcenter](https://github.com/elementary/appcenter/blob/e3a9b4feb4372621e53f96e56a6240c490548d43/src/Views/AppListUpdateView.vala). Bluetooth plug does not show a count of available devices and AppCenter always has the total count at the time the header is added.

_Originally posted by @meisenzahl in https://github.com/elementary/switchboard-plug-about/issues/195#issuecomment-774416260_

meisenzahl:
@danrabbit thanks so much for providing a solution! I had seen your PR unfortunately too late. I have now solved it in https://github.com/elementary/switchboard-plug-about/pull/195. I'm really sorry for wasting your precious time.

PR #199, opened 2021-02-08
Re-namespace Fwupd classes to About.Firmware

danrabbit:
If we want to use LibFwupd going forward, these namespaces will conflict


meisenzahl (dismissed):
LGTM. I renamed `FwupdManager.vala` to `FirmwareManager.vala`

Issue #200, opened 2021-02-09
Provide fallback icon for firmware devices


cassidyjames:
## What Happened

I looked at the firmware tab on my Dell Precision 5530 and noticed the Touchpad device has no icon, which looks weird in the list.

![Screenshot from 2021-02-09 11-30-30](https://user-images.githubusercontent.com/611168/107410444-8c9d7700-6aca-11eb-94de-66144dcbfb54.png)

## Expected Behavior

All devices have icons, even if it's a fallback firmware/device icon

## Steps to Reproduce

1. Have a Dell Precision 5530 touchpad or other device without a fallback icon
2. Open the Firmware page
3. See the Touchpad device with no icon

## Platform Information

elementary OS 6 Early Access, up to date

PR #201, opened 2021-02-09
Provide fallback icon for firmware devices

meisenzahl:
Closes https://github.com/elementary/switchboard-plug-about/issues/200


danrabbit (approved):
Yup, can confirm that this provides a fallback icon. Nice work

PR #202, opened 2021-02-09
Use LibFwupd

danrabbit:
Replace everything custom with LibFwupd


davidmhewitt:
This has gone back to introducing a noticeably significant amount delay when opening the plug, especially on slower hardware or when fwupd isn't running and has to be automatically started.

I can see that async methods aren't available in the vapi we have in the repos, but, after some digging, it looks like async methods are available in libfwupd, but only from around version 1.5 or 1.5.5, and we have 1.3.

The code cleanup is nice, but I don't know how we feel about (re-)introducing a bunch of UI latency that we just went to the trouble of fixing. The other option is to wrap these libfwupd methods in threads until we get a version that doesn't need them.

src/Widgets/FirmwareUpdateRow.vala
```diff
@@ -53,59 +56,107 @@ public class About.Widgets.FirmwareUpdateRow : Gtk.ListBoxRow {
...
-            case Firmware.ReleaseFlag.IS_UPGRADE:
-                if (device.latest_release.version == device.version) {
-                    break;
-                }
+        // TODO: Be smarter to avoid segfault
+        var icons = device.get_icons ();
+        if (icons != null) {
+            for (int i = 0; i < icons.length; i++) {
+                image.icon_name = icons[i];
+            }
```
tintou, line 68: something like this should work:
```suggestion
            image.icon = new GLib.ThemedIcon.from_names (icons);
```

danrabbit: It looks like it doesn't, because what we get from `get_icons` isn't necessarily a string array. So probably what needs to be done is to add only values which are strings to a new array and then use themedicon.from_names on that

meisenzahl:
> This has gone back to introducing a noticeably significant amount delay when opening the plug, especially on slower hardware or when fwupd isn't running and has to be automatically started.

I can confirm this..

> I can see that async methods aren't available in the vapi we have in the repos, but, after some digging, it looks like async methods are available in libfwupd, but only from version 1.5, and we have 1.3.
> 
> The code cleanup is nice, but I don't know how we feel about (re-)introducing a bunch of UI latency that we just went to the trouble of fixing. The other option is to wrap these libfwupd methods in threads until we get a version that doesn't need them.

@davidmhewitt what would be the preferred way?

meisenzahl:
@danrabbit were you still going to look at the details bit? Otherwise, I can create a PR for your PR on the weekend 🙂

danrabbit:
> @danrabbit were you still going to look at the details bit? Otherwise, I can create a PR for your PR on the weekend 🙂

I think it would be better to wait and just keep this scoped to reaching feature parity with master so that it doesn't get held up or become too complicated to review

danrabbit:
@davidmhewitt TIL how to GLib.Thread. Let me know if that doesn't fix it for you :)

@meisenzahl ah sorry I actually misunderstood your comment! I thought you were talking about doing the sliding details pane thing in this branch. I was able to get the details dialog working

davidmhewitt:
It's a lot less bad now, but I'm still seeing a bit of blocking on loading the plug.

All of the fwupd methods we have available to us are synchronous, so `get_devices`, `get_release_details`, `install`, etc...

I don't actually have any devices to update, but I assume the call to `install` is potentially quite long running and I also suspect that it it will freeze the UI entirely for the duration of that call.

As a general approach for making thing async using threads, you can do something like the following and then just use it like a regular async method (like the ones removed by this PR):

```
public async ReturnType example_method () {
    SourceFunc callback = example_method.callback;
    ReturnType result;
    
    new Thread<void> ("example_method", () => {
        // Pass back result and schedule callback
        result = long_running_synchronous_thing ();
        Idle.add ((owned) callback);
    });

    // Wait for background thread to schedule our callback
    yield;
    
    return result;
}
```

davidmhewitt:
Additionally, I forgot to mention, you shouldn't do anything that touches UI inside a thread. So your call to `update_list_view` inside your new thread is possibly dangerous.

Surrounding stuff with `Idle.add` passes it back to the main UI thread.

danrabbit:
@davidmhewitt Okay, I think I did it! Let me know if anything looks wrong

src/Views/FirmwareView.vala
```diff
@@ -76,14 +73,21 @@ public class About.FirmwareView : Gtk.Stack {
         add (grid);
         add (progress_view);
 
-        fwupd.on_device_added.connect (on_device_added);
-        fwupd.on_device_error.connect (on_device_error);
-        fwupd.on_device_removed.connect (on_device_removed);
+        try {
+            var fwupd_client = new Fwupd.Client ();
+            FirmwareClient.connect.begin (fwupd_client, (obj, res) => {
+                fwupd_client.device_added.connect (on_device_added);
```
davidmhewitt, line 79: The `try/catch` has to surround the `async_method.end` call which is missing here.

src/Views/FirmwareView.vala
```diff
@@ -92,67 +96,62 @@ public class About.FirmwareView : Gtk.Stack {
...
+                for (int i = 0; i < devices.length; i++) {
+                    add_device (client, devices[i]);
+                }
+
+                update_list.show_all ();
+            } catch (Error e) {
+                // TODO: Use placeholder
+                critical (e.message);
+            }
+        });
```
davidmhewitt, line 111: If you want to await the result of an async method inside an async method, you can use `yield` instead of `begin`/`end`. 

The original code that was removed here was:
```
 foreach (var device in yield fwupd.get_devices ()) {
    add_device (device);
}
```

That can pretty much be put back using your new method.

src/Widgets/FirmwareUpdateRow.vala
```diff
@@ -53,59 +56,121 @@ public class About.Widgets.FirmwareUpdateRow : Gtk.ListBoxRow {
...
         }
 
-        if ((yield fwupd.install (device, path)) == true) {
-            if (device.has_flag (Firmware.DeviceFlag.NEEDS_REBOOT)) {
-                show_reboot_dialog ();
-            } else if (device.has_flag (Firmware.DeviceFlag.NEEDS_SHUTDOWN)) {
-                show_shutdown_dialog ();
+        var path = yield download_file (release.get_uri ());
+
+        FirmwareClient.install.begin (client, device.get_id (), path, (obj, res) => {
```
davidmhewitt, line 116: Again, looks like you can use `yield` here.

davidmhewitt (changes requested):
A quick first glance review of your async changes

src/Widgets/FirmwareUpdateRow.vala
```diff
@@ -53,59 +56,116 @@ public class About.Widgets.FirmwareUpdateRow : Gtk.ListBoxRow {
...
+
 
         add (grid);
     }
 
-    private async void update (Firmware.Device device, Firmware.Release release) {
-        var path = yield fwupd.download_file (device, release.uri);
+    private async void update (Fwupd.Release release) {
+        var detach_caption = release.get_detach_caption ();
+        var detach_image = release.get_detach_image ();
```
tintou, line 101:
```suggestion
        unowned var detach_caption = release.get_detach_caption ();
        unowned var detach_image = release.get_detach_image ();
```

src/Widgets/FirmwareUpdateRow.vala
```diff
@@ -53,59 +56,116 @@ public class About.Widgets.FirmwareUpdateRow : Gtk.ListBoxRow {
...
-        var details = yield fwupd.get_release_details (device, path);
+        if (detach_image != null) {
+            detach_image = yield download_file (detach_image);
+        }
 
-        if (details.caption != null) {
-            if (show_details_dialog (details) == false) {
-                return;
-            }
+        if (detach_caption == null || show_details_dialog (detach_caption, detach_image) == false) {
```
davidmhewitt, line 107: This doesn't seem right? `detach_caption` is an optional string that gives the user instructions on how to put the device into the right mode for a firmware update. If it's omitted, we don't want to cancel the update by returning here?

davidmhewitt (changes requested):
I've just left one comment, but otherwise the code looks good. I don't have the means to test any of this though.

src/Widgets/FirmwareUpdateRow.vala
danrabbit: Ah yup, good catch. Mixed up the logic there. Should be fixed

src/Widgets/FirmwareUpdateRow.vala
```diff
@@ -53,59 +56,118 @@ public class About.Widgets.FirmwareUpdateRow : Gtk.ListBoxRow {
...
+        ) {
+            badge_icon = new ThemedIcon ("dialog-error"),
+            transient_for = (Gtk.Window) get_toplevel ()
+        };
+        message_dialog.show_all ();
+        message_dialog.run ();
+        message_dialog.destroy ();
+    }
+
+    private bool show_details_dialog (string detach_caption, string detach_image) {
```
davidmhewitt, line 166:
```suggestion
    private bool show_details_dialog (string detach_caption, string? detach_image) {
```

davidmhewitt (changes requested):

davidmhewitt (approved):
Code looks good. Don't have any updatable devices to test with.

PR #204, opened 2021-02-26
FirmwareView: subclass SimpleSettingsPage

### LINKED ISSUE #196: Firmware page needs to actually explain what a firmware is
## Problem

Everyone has heard what a firmware is and also likely knows it has something to do with computers and things that are firm, but the average person will not actually know what it really is.

## Proposal

elementary OS is not an IT guidebook but I see no reason to give the firmware page at least some text and maybe even an icon, in a visually consistent way to the Network and Bluetooth plugs.

## Prior Art

Screenshot from 2021-02-05 18-32-53

danrabbit:
Fixes #196 

Screenshot from 2021-02-26 10 07 47@2x


hanaral:
If there aren't any issues, it might be better to just merge something this incremental now and iterate later

cassidyjames (approved):

PR #205, opened 2021-03-07
Fix logic if device is up to date

meisenzahl:
The port to libfwupd has introduced a regression related to the logic of whether a device is up to date.

## Before

Bildschirmfoto von 2021-03-07 16 46 04

Bildschirmfoto von 2021-03-07 16 46 56

## After

Bildschirmfoto von 2021-03-07 16 56 53

In both cases, "Unifying Receiver" is at version RQR12.07_B0029 which is lower than the latest release RQR12.10_B0032.

### DIFF
--- a/src/Views/FirmwareView.vala
+++ b/src/Views/FirmwareView.vala
@@ -117,22 +117,36 @@ public class About.FirmwareView : Gtk.Stack {
 
     private void add_device (Fwupd.Client client, Fwupd.Device device) {
         if (device.has_flag (Fwupd.DEVICE_FLAG_UPDATABLE)) {
-            var row = new Widgets.FirmwareUpdateRow (client, device);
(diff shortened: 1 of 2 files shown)


meisenzahl:
@cassidyjames would be nice if you can try if this works for you too :smiley:

meisenzahl:
@elementary/ux should a device that can be updated show the current version of the firmware or the version that can be installed? :thinking:

cassidyjames:
I would think the available version, but I don't know—I could also see a format like `1.1.1 → 1.2.0` or something to make it more explicit, especially since we have plenty of room on that line.

meisenzahl:
Bildschirmfoto von 2021-03-09 07 11 14

cassidyjames (approved):

PR #206, opened 2021-03-10
Add detail page for firmware releases - Revenge of libfwupd

meisenzahl:
Supersedes https://github.com/elementary/switchboard-plug-about/pull/182
Closes https://github.com/elementary/switchboard-plug-about/issues/172

### DIFF
--- a/src/Views/FirmwareView.vala
+++ b/src/Views/FirmwareView.vala
@@ -28,6 +28,7 @@ public class About.FirmwareView : Gtk.Stack {
     private Gtk.Grid progress_view;
     private Gtk.ListBox update_list;
     private uint num_updates = 0;
+    private Fwupd.Client fwupd_client;
 
     construct {
         transition_type = Gtk.StackTransitionType.SLIDE_LEFT_RIGHT;
@@ -71,6 +72,7 @@ public class About.FirmwareView : Gtk.Stack {
         deck.add (update_scrolled);
 
         firmware_release_view = new FirmwareReleaseView ();
+        firmware_release_view.update.connect (on_update);
         deck.add (firmware_release_view);
 
         deck.visible_child = update_scrolled;
@@ -88,7 +90,7 @@ public class About.FirmwareView : Gtk.Stack {
         add (grid);
         add (progress_view);
 
-        var fwupd_client = new Fwupd.Client ();
+        fwupd_client = new Fwupd.Client ();
         FirmwareClient.connect.begin (fwupd_client, (obj, res) => {
             try {
                 FirmwareClient.connect.end (res);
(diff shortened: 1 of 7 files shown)


src/Views/FirmwareView.vala
```diff
@@ -180,6 +197,25 @@ public class About.FirmwareView : Gtk.Stack {
         update_list.show_all ();
     }
 
+    private void on_update (Fwupd.Device device, Fwupd.Release release) {
+        update_start (device);
+
+        update.begin (device, release, (obj, res) => {
+            update.end (res);
+            update_end ();
```
danrabbit, line 205: This doesn't appear to be used anywhere outside of this function. Probably not necessary to create a function for two lines right?

src/Views/FirmwareView.vala
```diff
@@ -180,6 +197,25 @@ public class About.FirmwareView : Gtk.Stack {
         update_list.show_all ();
     }
 
+    private void on_update (Fwupd.Device device, Fwupd.Release release) {
+        update_start (device);
```
danrabbit, line 201: This doesn't appear to be used anywhere outside of this function, either

src/Views/FirmwareView.vala
```diff
@@ -180,6 +197,25 @@ public class About.FirmwareView : Gtk.Stack {
         update_list.show_all ();
     }
 
+    private void on_update (Fwupd.Device device, Fwupd.Release release) {
```
danrabbit, line 200: That's kind of confusing to have a function `on_update` and a function `update`. Maybe one or both of these should be renamed for clarity on what they do

danrabbit (changes requested):
First pass here

danrabbit (changes requested):
I'm getting a segfault when the update directions dialog is being shown. Looks like an issue with getting the device icon:

```
** (io.elementary.switchboard:134065): WARNING **: 13:08:54.539: HardwareView.vala:433: Error opening file /sys/block/queue/rotational: No such file or directory

(io.elementary.switchboard:134065): Gtk-CRITICAL **: 13:09:01.662: gtk_widget_draw: assertion '!widget->priv->alloc_needed' failed

** (io.elementary.switchboard:134065): CRITICAL **: 13:09:01.741: granite_message_dialog_construct_with_image_from_icon_name: assertion 'image_icon_name != NULL' failed

(io.elementary.switchboard:134065): GLib-GObject-CRITICAL **: 13:09:01.741: g_object_ref_sink: assertion 'G_IS_OBJECT (object)' failed

(io.elementary.switchboard:134065): Gtk-CRITICAL **: 13:09:01.741: gtk_window_set_transient_for: assertion 'GTK_IS_WINDOW (window)' failed
Segmentation fault
```

danrabbit (approved):

PR #214, opened 2021-05-20
Handle long hardware names

JoseExposito:
- Use middle ellipsis in hardware names to make sure it's easy to distinguish the devices and also, in the case of audio devices, to be consistent with how the indicator displays them.

- Add margin to the left and right so long hardware names are properly displayed.

Before and after:

bug_before

bug_after

### DIFF
--- a/src/Views/HardwareView.vala
+++ b/src/Views/HardwareView.vala
@@ -45,33 +45,33 @@ public class About.HardwareView : Gtk.Grid {
         fetch_hardware_info ();
 
         var product_name_info = new Gtk.Label (Environment.get_host_name ()) {
-            ellipsize = Pango.EllipsizeMode.END,
+            ellipsize = Pango.EllipsizeMode.MIDDLE,
             selectable = true,
             xalign = 0
         };
         product_name_info.get_style_context ().add_class (Granite.STYLE_CLASS_H2_LABEL);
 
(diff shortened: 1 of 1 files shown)


src/Views/HardwareView.vala
```diff
@@ -152,6 +152,8 @@ public class About.HardwareView : Gtk.Grid {
             details_grid.add (manufacturer_website_info);
         }
 
+        margin_left = 32;
+        margin_right = 32;
```
cassidyjames, line 156: These should be 16 like the other views afaict

```suggestion
        margin_left = 16;
        margin_right = 16;
```

src/Views/HardwareView.vala
```diff
@@ -117,7 +117,7 @@ public class About.HardwareView : Gtk.Grid {
             }
 
             var manufacturer_info = new Gtk.Label (manufacturer_name) {
-                ellipsize = Pango.EllipsizeMode.END,
+                ellipsize = Pango.EllipsizeMode.MIDDLE,
                 selectable = true,
```
cassidyjames, line 121:
```suggestion
                selectable = true,
                tooltip_text = manufacturer_name,
```

cassidyjames (approved):
We can do the tooltips in a follow-up.

PR #218, opened 2021-05-29
Fallback from switcheroo to gnome session manager for GPU info.

### LINKED ISSUE #217: 11th-gen Intel graphics just show as "Intel"
I think this should be Intel Xe? I can look up the exact specs if it's helpful.

Screenshot from 2021-05-28 15-53-33@2x

vjr:
Fixes #217


vjr:
This is what I get now:

Screenshot from 2021-05-29 16 01 11

vjr:
Do I need to add a regex replacement to remove the "Mesa" bit? What about the end part (TGL GT2)?

vjr:
Removed the "Mesa" part and left the TGL GT2 part alone so that (some?) people can identify the model of GPU they have and not wonder what "Mesa" means.

vjr:
Looks like this now...

Screenshot from 2021-06-01 17 29 57

cassidyjames (approved):

PR #220, opened 2021-07-03
Spelling: -firmware updates

comradekingu:
Upgrades aren't updates, and upgrading firmware even less so.


src/Views/FirmwareView.vala
```diff
@@ -120,12 +120,12 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
...
 
-            placeholder_alert_view.title = _("Firmware Updates Are Not Available");
-            placeholder_alert_view.description = _("Firmware updates are not supported on this or any connected devices.");
+            placeholder_alert_view.title = _("No new firmware available");
+            placeholder_alert_view.description = _("None of your devices support installing new firmware this way.");
             update_list.show_all ();
         } catch (Error e) {
-            placeholder_alert_view.title = _("The Firmware Service Is Not Available");
-            placeholder_alert_view.description = _("Please make sure “fwupd” is installed and enabled.");
+            placeholder_alert_view.title = _("Firmware service unvailable");
```
jeremypw, line 127: Spelling: "unvailable! should be "unavailable"

src/Views/FirmwareView.vala
```diff
@@ -354,8 +354,8 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
 
     private void show_reboot_dialog () {
         var message_dialog = new Granite.MessageDialog.with_image_from_icon_name (
-            _("An update requires the system to restart to complete"),
-            _("This will close all open applications and restart this device."),
+            _("Installing new firmware requires a system reboot to complete"),
```
jeremypw, line 357: As the wingpanel indicator uses the words "Shutdown ... Restart", I do not see the advantage of using the technical word "reboot" instead.

comradekingu: Says "Reboot" in LXQt. Reboot is always start something up again in a booting fashion, whereas "restart" could be something just unloading and starting again.
Guess I am reverting this.

src/Views/FirmwareView.vala
comradekingu: Good catch!

comradekingu:
```suggestion
            placeholder_alert_view.title = _("Firmware service unavailable");
```

src/Views/FirmwareView.vala
```diff
@@ -33,7 +33,7 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
         Object (
             icon_name: "application-x-firmware",
             title: _("Firmware"),
-            description: _("Firmware updates provided by device manufacturers can improve performance and fix critical security issues.")
+            description: _("Any new firmware is provided by respective device manufacturers, in the hopes of improving performance and fixing security issues.")
```
danrabbit, line 36: This seems unnecessarily "lawyer-y" and I don't think improves clarity

src/Views/FirmwareView.vala
```diff
@@ -47,7 +47,7 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
 
         placeholder_alert_view = new Granite.Widgets.AlertView (
             _("Checking for Updates"),
-            _("Connecting to the firmware service and searching for updates."),
+            _("Connecting to the service to look for new firmware…"),
```
danrabbit, line 50: This sounds less natural to me. Your device already has some firmware, so it's not inaccurate to say that we're looking for updates

src/Views/FirmwareView.vala
```diff
@@ -120,12 +120,12 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
                 add_device (devices[i]);
             }
 
-            placeholder_alert_view.title = _("Firmware Updates Are Not Available");
-            placeholder_alert_view.description = _("Firmware updates are not supported on this or any connected devices.");
+            placeholder_alert_view.title = _("No new firmware available");
```
danrabbit, line 123: This is not necessarily true. There may be new firmware for your device, it's just not available in fwupd

src/Views/FirmwareView.vala
```diff
@@ -120,12 +120,12 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
...
+            placeholder_alert_view.description = _("None of your devices support installing new firmware this way.");
```
danrabbit, line 124: I like the attempt at using more natural language here, but "this way" seems vague and kind of implies an unanswered question of there being another way to install updates

src/Views/FirmwareView.vala
```diff
@@ -120,12 +120,12 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
...
             update_list.show_all ();
         } catch (Error e) {
-            placeholder_alert_view.title = _("The Firmware Service Is Not Available");
-            placeholder_alert_view.description = _("Please make sure “fwupd” is installed and enabled.");
+            placeholder_alert_view.title = _("Firmware service unavailable");
```
danrabbit, line 127: I don't have hugely strong feelings on the use of "unavailable" versus "not available", but I think it sounds more natural to say "The firmware service is unavailable" than to exclude these words

src/Views/FirmwareView.vala
```diff
@@ -216,7 +216,7 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
     private void header_rows (Widgets.FirmwareUpdateRow row1, Widgets.FirmwareUpdateRow? row2) {
         if (row2 == null && row1.is_updatable) {
             var header = new FirmwareHeaderRow (
-                dngettext (GETTEXT_PACKAGE, "%u Update Available", "%u Updates Available", num_updates).printf (num_updates)
+                dngettext (GETTEXT_PACKAGE, "%u new firmware available", "%u new firmwares available", num_updates).printf (num_updates)
```
danrabbit, line 219: This sounds much less natural. In US English I don't believe that firmware uses discrete units. It sounds like saying "10 waters". This also should be title case since it is a row header

src/Views/FirmwareView.vala
```diff
@@ -240,7 +240,7 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
     }
 
     private async void update (Fwupd.Device device, Fwupd.Release release) {
-        progress_alert_view.title = _("“%s” is being updated").printf (device.get_name ());
+        progress_alert_view.title = _("“%s” is getting new firmware…").printf (device.get_name ());
```
danrabbit, line 243: This sounds much less natural

src/Views/FirmwareView.vala
```diff
@@ -327,7 +327,7 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
         }
 
         var message_dialog = new Granite.MessageDialog (
-            _("“%s” needs to manually be put in update mode").printf (device.get_name ()),
+            _("“%s” needs to be put in firmware-upgradable mode manually").printf (device.get_name ()),
```
danrabbit, line 330: I'm not sure about `firmware-upgradable mode`. I think more likely user manuals will say `update mode`. Otherwise, I agree it sounds more natural to swap the order of these things in the sentence

src/Views/FirmwareView.vala
danrabbit: I guess it's a good thing we use Pantheon and not LXQt :) We should definitely be internally consistent here and use "Restart"

src/Views/FirmwareView.vala
```diff
@@ -376,8 +376,8 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
 
     private void show_shutdown_dialog () {
         var message_dialog = new Granite.MessageDialog.with_image_from_icon_name (
-            _("An update requires the system to shut down to complete"),
-            _("This will close all open applications and turn off this device."),
+            _("Installing new firmware requires the system to be turned off to complete"),
```
danrabbit, line 379: I think we should probably stick with "shut down" since that's the name of the menu item in the power menu so it's more explicitly linked to something you may be familiar with. I agree that "turned off" sounds more natural, but in this specific case it might be slightly less clear

danrabbit (changes requested):
Thanks for working through the strings here! I appreciate the attention to detail. There are some good suggestions in here, but I think some of these sound a bit less natural or maybe less like US English. I've left a few inline comments

src/Views/FirmwareView.vala
comradekingu: The original opens a whole unanswered question of "what else can it do?"
I really want critical security issues to get fixed, yes. I don't want "can" to be the operative word in that regard.

src/Views/FirmwareView.vala
comradekingu: I'll simplify it to what actually happens.
Originally I thought to just add the ellipsis, but firmware upgrades are not updates.

src/Views/FirmwareView.vala
comradekingu: It also explains the process of what is actually happening.

src/Views/FirmwareView.vala
danrabbit: It really depends on the device manufacturer. Sometimes firmware updates enable new features or are just bug fixes. It is indeed open ended what the update may or may not include

src/Views/FirmwareView.vala
danrabbit: I'm curious what you feel is the difference between "upgrade" and "update"?

src/Views/FirmwareView.vala
comradekingu: Differs between products. It is no good looking for the technically dubious "update mode" when actually wanting "Turn on AHCI support" for example.

src/Views/FirmwareView.vala
comradekingu: I was going with the concept of "ware" many "wares", but that is quirky.

src/Views/FirmwareView.vala
comradekingu: I think the minimal amount of text is what I want when trying to keep on top of what is very stressful and possibly non-reversible. Title case for a sentence looks off.

src/Views/FirmwareView.vala
comradekingu: Yes, you can connect to pins and do it the old way.
Will change.

src/Views/FirmwareView.vala
comradekingu: Fair point, but the distinction isn't there between new and old.

comradekingu:
```suggestion
            placeholder_alert_view.title = _("No new firmware found");
```

src/Views/FirmwareView.vala
comradekingu:
```suggestion
            placeholder_alert_view.description = _("Nothing on or connected to your system supports installing new firmware this way.");
```

src/Views/FirmwareView.vala
```diff
@@ -216,7 +215,7 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
     private void header_rows (Widgets.FirmwareUpdateRow row1, Widgets.FirmwareUpdateRow? row2) {
         if (row2 == null && row1.is_updatable) {
             var header = new FirmwareHeaderRow (
-                dngettext (GETTEXT_PACKAGE, "%u Update Available", "%u Updates Available", num_updates).printf (num_updates)
+                dngettext (GETTEXT_PACKAGE, "%u New", "%u New", num_updates).printf (num_updates)
```
danrabbit, line 218: This is not more clear. "1 New" what? These are available updates

src/Views/FirmwareView.vala
```diff
@@ -33,7 +33,7 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
         Object (
             icon_name: "application-x-firmware",
             title: _("Firmware"),
-            description: _("Firmware updates provided by device manufacturers can improve performance and fix critical security issues.")
+            description: _("New firmware is released by device manufacturers to improve performance and fix security issues.")
```
comradekingu, line 36:
```suggestion
            description: _("New firmware is released by device manufacturers to improve performance, and fix security issues.")
```

src/Views/FirmwareView.vala
```diff
@@ -46,8 +46,7 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
         progress_alert_view.get_style_context ().remove_class (Gtk.STYLE_CLASS_VIEW);
 
         placeholder_alert_view = new Granite.Widgets.AlertView (
-            _("Checking for Updates"),
-            _("Connecting to the firmware service and searching for updates."),
```
comradekingu, line 50: Is it OK to just drop this?

src/Views/FirmwareView.vala
```diff
@@ -354,8 +353,8 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
 
     private void show_reboot_dialog () {
         var message_dialog = new Granite.MessageDialog.with_image_from_icon_name (
-            _("An update requires the system to restart to complete"),
-            _("This will close all open applications and restart this device."),
+            _("Installing new firmware requires restarting the system"),
```
comradekingu, line 356:
```suggestion
            _("Restart the system to finish installing the new firmware"),
```

danrabbit (changes requested):
I'm not sure why you're avoiding the use of the word "update" here. This makes the copy much less clear. Users are coming here to install firmware updates. The word "update" here is correct.

src/Views/FirmwareView.vala
```diff
@@ -376,8 +375,8 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
 
     private void show_shutdown_dialog () {
         var message_dialog = new Granite.MessageDialog.with_image_from_icon_name (
-            _("An update requires the system to shut down to complete"),
-            _("This will close all open applications and turn off this device."),
+            _("The system must be shut down after installing new firmware"),
```
comradekingu, line 378:
```suggestion
            _("Shut down the system to finish installing the new firmware"),
```

src/Views/FirmwareView.vala
comradekingu: I imagined this was under some other title since the original is just "update".

comradekingu:
```suggestion
                dngettext (GETTEXT_PACKAGE, "%u available firmware upgrade", "%u available firmware upgrades", num_updates).printf (num_updates)
```

comradekingu: I am a fan of "One available", but don't know if that causes issues for any translators.

src/Views/FirmwareView.vala
```diff
@@ -216,7 +215,7 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
     private void header_rows (Widgets.FirmwareUpdateRow row1, Widgets.FirmwareUpdateRow? row2) {
         if (row2 == null && row1.is_updatable) {
             var header = new FirmwareHeaderRow (
-                dngettext (GETTEXT_PACKAGE, "%u Update Available", "%u Updates Available", num_updates).printf (num_updates)
+                dngettext (GETTEXT_PACKAGE, "%u available firmware upgrade", "%u available firmware upgrades", num_updates).printf (num_updates)
```
comradekingu, line 218:
```suggestion
                dngettext (GETTEXT_PACKAGE, "One available firmware upgrade", "%u available firmware upgrades", num_updates).printf (num_updates)
```
Not sure if this causes issues for any translators.

src/Views/FirmwareView.vala
danrabbit: This is used in the UI as a section header, so it is intended to be title case. Since this whole pane deals with firmware, we don't need to specify that these are firmware updates. I don't think changing this string makes anything more clear

src/Views/FirmwareView.vala
danrabbit: This isn't the correct way to use ngettext. It needs to stick with the `%u` placeholder

src/Views/FirmwareView.vala
comradekingu:
```suggestion
            placeholder_alert_view.title = _("Could not connect to firmware service");
```

src/Views/FirmwareView.vala
```diff
@@ -120,12 +119,12 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
...
 
-            placeholder_alert_view.title = _("Firmware Updates Are Not Available");
-            placeholder_alert_view.description = _("Firmware updates are not supported on this or any connected devices.");
+            placeholder_alert_view.title = _("No new firmware found");
+            placeholder_alert_view.description = _("Nothing on or connected to your system supports installing new firmware this way.");
             update_list.show_all ();
         } catch (Error e) {
-            placeholder_alert_view.title = _("The Firmware Service Is Not Available");
-            placeholder_alert_view.description = _("Please make sure “fwupd” is installed and enabled.");
+            placeholder_alert_view.title = _("Could not connect to firmware service");
```
comradekingu, line 126:
```suggestion
            placeholder_alert_view.title = _("Could not connect to the firmware service");
```

src/Views/FirmwareView.vala
danrabbit: No, the alertview needs to have two lines or it won't compile

src/Views/FirmwareView.vala
```diff
@@ -120,12 +119,12 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
...
 
-            placeholder_alert_view.title = _("Firmware Updates Are Not Available");
-            placeholder_alert_view.description = _("Firmware updates are not supported on this or any connected devices.");
+            placeholder_alert_view.title = _("No new firmware found");
+            placeholder_alert_view.description = _("Nothing on or connected to your system supports installing new firmware this way.");
             update_list.show_all ();
         } catch (Error e) {
-            placeholder_alert_view.title = _("The Firmware Service Is Not Available");
-            placeholder_alert_view.description = _("Please make sure “fwupd” is installed and enabled.");
+            placeholder_alert_view.title = _("Could not connect to the firmware service");
```
comradekingu, line 126:
```suggestion
            placeholder_alert_view.title = _("Firmware service unavailable");
```

src/Views/FirmwareView.vala
```diff
@@ -33,7 +33,7 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
         Object (
             icon_name: "application-x-firmware",
             title: _("Firmware"),
-            description: _("Firmware updates provided by device manufacturers can improve performance and fix critical security issues.")
+            description: _("New firmware is released by device manufacturers to improve performance, and fix security issues.")
```
danrabbit, line 36: "fix security issues" isn't an independent clause, it's part of a list, so it doesn't make sense to add a comma here. I'm unconvinced that this adds a significant amount of clarity that would make it worth it to lose existing translations

src/Views/FirmwareView.vala
```diff
@@ -120,12 +119,12 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
                 add_device (devices[i]);
             }
 
-            placeholder_alert_view.title = _("Firmware Updates Are Not Available");
-            placeholder_alert_view.description = _("Firmware updates are not supported on this or any connected devices.");
+            placeholder_alert_view.title = _("No new firmware found");
+            placeholder_alert_view.description = _("Nothing on or connected to your system supports installing new firmware this way.");
```
danrabbit, line 123: I don't think this is a better string than what is already there. It doesn't seem to clarify anything and instead adds a new question of where else firmware updates can be installed from

danrabbit (changes requested):
I think all of the instances where you've replaced "updates" with "new firmware" have either made no noticeable improvement or decreased clarity about what's happening. It's not an improvement to make that change

src/Views/FirmwareView.vala
comradekingu: To me that Oxford comma makes sense, because it isn't always both.
Translations are only fuzzified. Translators don't like vague sentences at all, because it introduces active errors or even vaguer language (without a lot of effort).

src/Views/FirmwareView.vala
comradekingu: It denotes other venues of upgrading firmware outside of the tool are viable, as it should.
(Ideally the tool should alert you to this if need be, but that is something for fwupd to fix.)
Unfortunately there is no peace of mind in relying on only one solution here. It is also unsolvable, so that is going to remain the case.

Newer firmware being available does not depend on fwupd being able to find or upgrade it, so the statement is factually untrue.

src/Views/FirmwareView.vala
comradekingu: And the second can't be empty even?

src/Views/FirmwareView.vala
danrabbit: No the comma is definitely incorrect here. The Oxford comma would come into play if you were listing 3 or more items.

The suggestion here is not necessarily always true though. You're trying to make a definitive statement, but that makes the statement misleading. It's important to update firmware because it could contain these kinds of fixes that are important to users, but it could just as easily contain neither of these things. Saying that firmware updates can contain these kinds of fixes instead of saying they for sure do contain these types of fixes is intentional

src/Views/FirmwareView.vala
danrabbit: It's not designed to be empty, it would not be spaced or aligned correctly

src/Views/FirmwareView.vala
danrabbit: It's not within the scope of this UI to tell people to sideload their own firmware, and that's probably not advisable for most people.

The UI is referring to the systems ability to update firmware and that firmware updates through the system are not supported. This is why it is correct and important that we are talking about firmware updates and not just "new firmware"

comradekingu: Neither option tells anyone to install firmware manually.
The original says firmware "updates" aren't available, because they aren't supported, which is a mixed message, and often also wrong.
What the user with should do beyond knowing the tool is not exhaustive is up to the user.

What is "new firmware" if not firmware upgrades?

src/Views/FirmwareView.vala
comradekingu: True, it sometimes introduces a backdoor. I was thinking that if the user is there to do something, the why of why it should be done is the important bit to convey. Will try to think of something that works.
I think explaining "installed on actual hardware" should maybe be here instead of the looking for fw bit.


It isn't incorrect, and it removes ambiguity. That is the point of an Oxford comma. 3 or more is a serialized list, the point is the relation between them, also as it relates to the former or latter.
If for no other reason than you could have added one in "No the comma is definitely incorrect here."

src/Views/FirmwareView.vala
comradekingu: Nice to know :)

src/Views/FirmwareView.vala
comradekingu: Good to know. Will it then display as "1" or "One"?
I suppose a way to avoid it is to use a developer locale and then translate it in en_UK and en_US and so on as "One"?
I have been using "Én" in nb_NO since that is the more correct way to do it.

src/Views/FirmwareView.vala
```diff
@@ -33,7 +33,7 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
         Object (
             icon_name: "application-x-firmware",
             title: _("Firmware"),
-            description: _("Firmware updates provided by device manufacturers can improve performance and fix critical security issues.")
+            description: _("Upgrade internal systems stored or running on system- and pheriperial devices issued by their makers, which is sometimes critical for performance and security.")
```
comradekingu, line 36: A bit longer now than "New firmware is released by device manufacturers to improve performance, and fix security issues." but all bases are at least covered.

src/Views/FirmwareView.vala
```diff
@@ -46,8 +46,8 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
         progress_alert_view.get_style_context ().remove_class (Gtk.STYLE_CLASS_VIEW);
 
         placeholder_alert_view = new Granite.Widgets.AlertView (
-            _("Checking for Updates"),
-            _("Connecting to the firmware service and searching for updates."),
+            _("Finding new firmware…"),
+            _("Looking for software to be installed on your actual hardware…"),
```
comradekingu, line 50:
```suggestion
            _("Looking for software to be installed on your hardware…"),
```

comradekingu:
```suggestion
            _("Looking for new software for your hardware…"),
```

src/Views/FirmwareView.vala
```diff
@@ -46,8 +46,8 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
         progress_alert_view.get_style_context ().remove_class (Gtk.STYLE_CLASS_VIEW);
 
         placeholder_alert_view = new Granite.Widgets.AlertView (
-            _("Checking for Updates"),
-            _("Connecting to the firmware service and searching for updates."),
+            _("Finding new firmware…"),
+            _("Looking for software to be installed on your hardware…"),
```
comradekingu, line 50:
```suggestion
            _("Looking for new software for your hardware…"),
```

src/Views/FirmwareView.vala
```diff
@@ -46,8 +46,8 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
         progress_alert_view.get_style_context ().remove_class (Gtk.STYLE_CLASS_VIEW);
 
         placeholder_alert_view = new Granite.Widgets.AlertView (
-            _("Checking for Updates"),
-            _("Connecting to the firmware service and searching for updates."),
+            _("Finding new firmware…"),
+            _("Looking for new software for your hardware…"),
```
comradekingu, line 50: I feel this string strikes the right trade-off.

src/Views/FirmwareView.vala
danrabbit: In US English it will display as "1", which is intended for that locale

src/Views/FirmwareView.vala
comradekingu:
```suggestion
            description: _("Upgrade internal systems stored or running on system- and peripheral devices (issued by their makers), sometimes containing critical performance and security fixes.")
```

src/Views/FirmwareView.vala
danrabbit: In US English it will display as "1", which is intended for that locale

comradekingu: Maybe the jury is out on that https://www.grammarly.com/blog/when-to-spell-out-numbers/ (this seems to be my position).
In the very least you would need a list of genders to pick the right one (intended ;) word to spell it out with in other languages.

cassidyjames:
This PR makes a ton of string changes that all are individually contested, and now conflicts with master. I'm going to close this, but feel free to open smaller PRs that change fewer strings so those can be evaluated and potentially approved on their own.

src/Views/FirmwareView.vala
comradekingu: Update to pull in the list of packages, and upgrade to install those.
I made an effort to avoid it all though.

comradekingu:
@cassidyjames 
It has to be a consistent approach, or it is worthless.
This is neither professional or consistent in its own right, and it breaks with conventions used elsewhere in elementary OS.
I can detail some errors made and the confusion that arises, from an electronics background, in translation, or support point of view.

You both failed to see it, and that means users suffer.
There is just one contested string, and on the basis of "It doesn't seem to clarify anything …" you will sadly be proven wrong.
"It's not within the scope of this UI to tell people to sideload their own firmware" misses the point.
Vendors will have all sorts of other schemes to upgrade firmware, and that is safety-critical.
Nobody is being told how, but rather made aware.

src/Views/FirmwareView.vala
```diff
@@ -33,7 +33,7 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
         Object (
             icon_name: "application-x-firmware",
             title: _("Firmware"),
-            description: _("Firmware updates provided by device manufacturers can improve performance and fix critical security issues.")
```
comradekingu, line 36: They are not updates, and they can introduce critical safety issues and hamper performance.

src/Views/FirmwareView.vala
```diff
@@ -46,8 +46,8 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
         progress_alert_view.get_style_context ().remove_class (Gtk.STYLE_CLASS_VIEW);
 
         placeholder_alert_view = new Granite.Widgets.AlertView (
-            _("Checking for Updates"),
-            _("Connecting to the firmware service and searching for updates."),
```
comradekingu, line 50: There is no ending ellipsis here. "The firmware service" is needless complexion telling the user nothing of value.
Explaining the connection made to it makes no sense, because it is just a technical aspect without real distinction.
What are "updates"? We deduce it has something to do with "firmware", which is not explained.

src/Views/FirmwareView.vala
```diff
@@ -376,8 +376,8 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
 
     private void show_shutdown_dialog () {
         var message_dialog = new Granite.MessageDialog.with_image_from_icon_name (
-            _("An update requires the system to shut down to complete"),
-            _("This will close all open applications and turn off this device."),
```
comradekingu, line 380: "will" is redundant.

src/Views/FirmwareView.vala
```diff
@@ -376,8 +376,8 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
...
+            _("Shut down the system to finish installing the new firmware"),
+            _("This closes all open apps and turns off this device."),
```
comradekingu, line 380: Should be "the device" to avoid the double _this_.

src/Views/FirmwareView.vala
```diff
@@ -354,8 +354,8 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
 
     private void show_reboot_dialog () {
         var message_dialog = new Granite.MessageDialog.with_image_from_icon_name (
-            _("An update requires the system to restart to complete"),
```
comradekingu, line 357: Double _to_.
To complete what? Left in the air.
The whole sentence is turned on its head, because the premise is that the system is required to restart, but only at the last words is it revealed to be "to complete". This again references "An update" all the way at the beginning.
It is taxing to read this kind of spaghetti-language, and it is made worse yet in translation.

src/Views/FirmwareView.vala
```diff
@@ -327,7 +327,7 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
         }
 
         var message_dialog = new Granite.MessageDialog (
-            _("“%s” needs to manually be put in update mode").printf (device.get_name ()),
```
comradekingu, line 330: There is no such thing as an "update mode". What that does is not explained.

src/Views/FirmwareView.vala
```diff
@@ -291,7 +291,7 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
         }
 
         if (!result) {
-            show_error_dialog (device, "Download of %s was not succesfull".printf (uri));
```
comradekingu, line 294: "Download of %s" is a semantic error. "not successful" should be "unsuccessful", and "successful" is superfluous.
Nowhere else in elementary OS is "successful/ly" used.

src/Views/FirmwareView.vala
```diff
@@ -240,7 +240,7 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
     }
 
     private async void update (Fwupd.Device device, Fwupd.Release release) {
-        progress_alert_view.title = _("“%s” is being updated").printf (device.get_name ());
```
comradekingu, line 243: This is the wrong tense, and it doesn't have an ellipsis.

src/Views/FirmwareView.vala
```diff
@@ -120,12 +120,12 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
...
             }
 
-            placeholder_alert_view.title = _("Firmware Updates Are Not Available");
-            placeholder_alert_view.description = _("Firmware updates are not supported on this or any connected devices.");
+            placeholder_alert_view.title = _("No new firmware found");
+            placeholder_alert_view.description = _("Nothing on or connected to your system supports installing new firmware this way.");
             update_list.show_all ();
         } catch (Error e) {
-            placeholder_alert_view.title = _("The Firmware Service Is Not Available");
-            placeholder_alert_view.description = _("Please make sure “fwupd” is installed and enabled.");
```
comradekingu, line 128: Shouldn't be "enabled", because the distinction to "turned on/off" is not meaningful here. (extra complexity invoked).
What "enabled" even is isn't explained. This means the user doesn't get to know what "fwupd" is.

src/Views/FirmwareView.vala
```diff
@@ -120,12 +120,12 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
                 add_device (devices[i]);
             }
 
-            placeholder_alert_view.title = _("Firmware Updates Are Not Available");
-            placeholder_alert_view.description = _("Firmware updates are not supported on this or any connected devices.");
+            placeholder_alert_view.title = _("No new firmware found");
+            placeholder_alert_view.description = _("Nothing on or connected to your system supports installing new firmware this way.");
             update_list.show_all ();
         } catch (Error e) {
-            placeholder_alert_view.title = _("The Firmware Service Is Not Available");
```
comradekingu, line 127: Sentence in title case. "not available" is inelegant.

src/Views/FirmwareView.vala
```diff
@@ -120,12 +120,12 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
                 add_device (devices[i]);
             }
 
-            placeholder_alert_view.title = _("Firmware Updates Are Not Available");
```
comradekingu, line 123: Sentence in title casing. Weird sentence at that.
Are they anything but unavailable?

src/Views/FirmwareView.vala
```diff
@@ -120,12 +120,12 @@ public class About.FirmwareView : Granite.SimpleSettingsPage {
...
-            placeholder_alert_view.description = _("Firmware updates are not supported on this or any connected devices.");
```
comradekingu, line 124: This has no merit. There could be any number of other means to upgrade the firmware.
This will remain the case so long as interfaces don't disclose the device or firmware version at all, or truthfully, which is to say for the foreseeable future..

comradekingu:
@danrabbit 
> I think some of these sound a bit less natural or maybe less like US English. I've left a few inline comments

I detailed the problems found, which mostly explains my choices. In my head there is no possible way a concern for what sounds more American could be paramount. Even so, can we work within the realms of gains made?

src/Views/FirmwareView.vala
cassidyjames: This sentence is to inform that firmware updates:
- are provided by device manufacturers (e.g. not elementary, and not a random third party)
- can improve performance
- can fix critical security issues

We can't guarantee that _all_ updates to firmware _definitely_ do those things, hence the intentional "can" language. However, communicating the potential  benefits of keeping your firmware up to date is important and the need this serves.

comradekingu: It just says "firmware" twice. That doesn't help people that don't already know what firmware is.
What you then get is _of the ones that are provided by_ the flowchart goes into all sorts of directions.
There is no knowing what the opposite for example ~"by elementary OS" does, when the sentence plays out as if getting FW upgrades from "device manufacturers" is the only reason it is qualified by "can improve ___"
"Device manufacturers" isn't clear at all. Absolutely nobody that doesn't already know better will think it also includes stuff on the actual computer. 

"Can" is not the same as "may".
Is it limited to those by device manufactures?
Those kinds of questions are the ones users are faced with when it is written this way, and it is almost every string.

I didn't remove the description, but instead attempted to explain what firwmare is and how it can be upgraded.
There might be a better explanation, but let's give people a fighting chance of understanding how their systems work and what they are doing.

PR #227, opened 2021-08-10
Use Pretty Hostname

### LINKED ISSUE #226: Use Pretty Hostname
## Problem

It looks like we're getting the value for static hostname which doesn't include spaces or capital letters etc

## Proposal

There's also a "Pretty Hostname" value that can contain spaces and such, so we should probably display that if it's available. This would give us a name like "Daniel's Yoga" instead of "daniels-yoga"

meisenzahl:
Closes #226

### DIFF
--- a/src/Views/HardwareView.vala
+++ b/src/Views/HardwareView.vala
@@ -44,7 +44,7 @@ public class About.HardwareView : Gtk.Grid {
     construct {
         fetch_hardware_info ();
 
-        var product_name_info = new Gtk.Label (Environment.get_host_name ()) {
+        var product_name_info = new Gtk.Label (get_host_name ()) {
             ellipsize = Pango.EllipsizeMode.MIDDLE,
             selectable = true,
             xalign = 0
@@ -487,6 +487,10 @@ public class About.HardwareView : Gtk.Grid {
     struct GraphicsReplaceStrings {
         string regex;
         string replacement;
+    }
+
+    private string get_host_name () {
+        return GLib.Environment.get_host_name ();
     }
 }
 


src/Views/HardwareView.vala
```diff
@@ -162,17 +162,8 @@ public class About.HardwareView : Gtk.Grid {
...
-                "org.freedesktop.hostname1",
-                "/org/freedesktop/hostname1"
-            );
-
-            manufacturer_logo.icon_name = system_interface.icon_name;
-        } catch (IOError e) {
-            critical (e.message);
-        }
+        get_system_interface_instance ();
+        manufacturer_logo.icon_name = system_interface.icon_name;
```
danrabbit, line 166: We should probably still have a null check here in case `get_system_interface_instance ()` fails for some reason

src/Views/HardwareView.vala
```diff
@@ -488,12 +479,41 @@ public class About.HardwareView : Gtk.Grid {
...
+            } catch (GLib.Error e) {
+                warning ("%s", e.message);
+            }
+        }
+    }
+
+    private string get_host_name () {
+        get_system_interface_instance ();
+
+        string hostname = system_interface.pretty_hostname;
```
danrabbit, line 500: Same here with the null check and we could fall back to Environment.get_host_name () in case `system_interface` is null

danrabbit (changes requested):
Seems to work fine here with fallbacks so I imagine we could merge this even before finishing the other branch.

Left a couple comments that we should probably still include null checks just in case we fail to connect to the system interface for some reason

danrabbit (approved):
Nice work!

PR #238, opened 2021-12-22
Avoid ugly (null) label on some rolling releases distributions

### LINKED ISSUE #237: On some rolling releases the OS version reads "(null)"
### What Happened?

When opening about section about my operating system (Archlinux, which is a rolling release), the OS name reads "Arch Linux (null)"
Capture du 2021-12-22 19-24-01

### Steps to Reproduce

1. Use a rolling release distribution (Archlinux…)
2. Install Pantheon desktop and particularly switchboard and switchboard-plug-about
3. Open About section of switchboard to read information about your operating system.

### Expected Behavior

When no version information are available, nothing should be shown.

Capture du 2021-12-22 19-27-04

### OS Version

Other Linux

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

milouse:
With the patch attached, the "(null)" string interpolation of missing OS version is removed from the operating system name label.

With that patch applied, the about window looks like this:

Capture du 2021-12-22 19-27-04

This fix #237

### DIFF
--- a/src/Views/OperatingSystemView.vala
+++ b/src/Views/OperatingSystemView.vala
@@ -72,7 +72,7 @@ public class About.OperatingSystemView : Gtk.Grid {
         // want more granular control over text formatting
         var pretty_name = "<b>%s</b> %s".printf (
             Environment.get_os_info (GLib.OsInfoKey.NAME),
-            Environment.get_os_info (GLib.OsInfoKey.VERSION)
+            Environment.get_os_info (GLib.OsInfoKey.VERSION) ?? ""
         );
 
         var title = new Gtk.Label (pretty_name) {


src/Views/OperatingSystemView.vala
```diff
@@ -70,9 +70,13 @@ public class About.OperatingSystemView : Gtk.Grid {
...
         // Intentionally not using GLib.OsInfoKey.PRETTY_NAME here because we
         // want more granular control over text formatting
+        var version_name = Environment.get_os_info (GLib.OsInfoKey.VERSION);
+        if (version_name == null) {
+            version_name = "";
+        }
         var pretty_name = "<b>%s</b> %s".printf (
             Environment.get_os_info (GLib.OsInfoKey.NAME),
-            Environment.get_os_info (GLib.OsInfoKey.VERSION)
+            version_name
```
meisenzahl, line 79:
```suggestion
            Environment.get_os_info (GLib.OsInfoKey.VERSION) ?? ""
```

You can simplify this by using the [null coalescing operator](https://naaando.gitbooks.io/the-vala-tutorial/content/en/3-basics/operators.html):

> `a ?? b` is equivalent to `a != null ? a : b`. This operator is useful for example to provide a default value in case a reference is _null_:

> ```csharp
> stdout.printf("Hello, %s!\n", name ?? "unknown person");
> ```

milouse: Thank you very much for you review and advice. I’ve integrated them :)

meisenzahl (approved):

PR #250, opened 2022-09-28
Handle Ryzen CPU and AMD GPU vendor string and break on first regex match

vjr:
Some fixups to handle specific Ryzen CPU and AMD GPU vendor strings in the regex replacement cleanup method.

Also, break after the first regex match, so the replacement strings are in order of preference now.

Before: (note the lowercasing of the cpu string and lengthy gpu string)

about_before

After: (note the generic gpu string part matched, instead of the "RX 570" trailer because the actual gpu is an RX 580)

about_after

### DIFF
--- a/src/Views/HardwareView.vala
+++ b/src/Views/HardwareView.vala
@@ -257,7 +257,7 @@ public class About.HardwareView : Gtk.Grid {
             } else if (cpu.@value == 6) {
                 result += _("Hexa-Core %s").printf (clean_name (cpu.key));
             } else {
-                result += "%u\u00D7 %s ".printf (cpu.@value, clean_name (cpu.key));
+                result += "%u \u00D7 %s ".printf (cpu.@value, clean_name (cpu.key));
             }
         }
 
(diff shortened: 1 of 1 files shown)


vjr:
Do we want to have "AMD®" (the registered trademark symbol) prefixed on the GPU (and CPU?) vendor string displayed?

lenemter (approved):
Works for me

Before:
image
After:
image

vjr:
@lenemter  I update the PR with an addition regex - could you see if it cleans up your iGPU vendor string? (post updated screenshot)?

lenemter (approved):
Thanks! Looks a lot better now

image

lenemter:
® is too small in this font, should we keep it?

vjr:
Thanks @lenemter ! I will let others (core devs) comment/merge as they see fit :-)

vjr:
> ® is too small in this font, should we keep it?

Seems OK to me (there is in existing code as well) but I don't have opinion either way.

ryonakano (approved):
Code looks good and works as expected for me.

# Before
スクリーンショット 2022-10-01 09 11 52

# After
スクリーンショット 2022-10-01 09 12 35

PR #252, opened 2022-10-04
Custom layperson-friendly format size

vjr:
The `GLib.format_size ()` function returns a single decimal place such as `16.0 GiB` (for memory) and `512.1 GB` (for storage) while this PR returns `16 GB` and `512 GB` instead, with the intent of being user-friendly for the layperson. Returns "GB" instead of "GiB" (for example) even if IEC_UNITS is requested.

Before:

about_after2

After:

about_after3


vjr:
I'm not sure if anything further needs to be done for the newly added translate strings?

tintou:
I'm -1 on this, we have to use the right unit.

tintou:
From the GLib documentation:
```
use IEC (base 1024) units with "KiB"-style suffixes. IEC units should only be used for reporting things with a strong "power of 2" basis, like RAM sizes or RAID stripe sizes. Network and storage sizes should be reported in the normal SI units.
```

tintou:
If people don't know the difference I feel like they can search on their own if they are really interested in this :)

I'm sure that if they don't understand it, they are anyway seeing the unit as a opaque thing so there is no difference between MiB and MB for them :upside_down_face:

vjr:
@tintou  how does it look now? (both code and screenshot?)

about_after4

vjr:
> If people don't know the difference I feel like they can search on their own if they are really interested in this :)
> 
> I'm sure that if they don't understand it, they are anyway seeing the unit as a opaque thing so there is no difference between MiB and MB for them upside_down_face

Personally I don't agree with "no difference between SI and IEC for ordinary users" - I think (I guess) average user knows what "MB" and "GB" are but not "MiB" and "GiB" which is why I prefer showing SI units for the memory label too, but not a strong opinion.

vjr:
I just looked at my Windows and Android devices and they show the memory capacity unit as GB and not GiB even though the value is of course IEC.

I searched on Google for screenshots of system information sections on macOS and iOS and there too it's GB instead of GiB.

I know this is a nitpick PR to propose but whenever I look at the current hardware view in OS 6.1 showing "15.7 GiB memory" instead of the proposed "16 GB memory" it triggers some sort of OCD in me :-P

avojak:
Just my 2¢ but I think it’s a good change. I can’t think of the last time I saw a computer advertised using GiB or RAM. I don’t have any evidence at all to back this up, but I think a user would be more confused seeing 15.7 GiB and wondering what happened to the other 0.3. I don’t like the argument that a user should go look up the difference themselves. Is GB the most technically accurate? No, but it doesn’t need to be in this case. In this context, I would expect my system to report in a format that matches how it would be advertised or generally discussed. If someone really cares that much, there are other system utilities to find exact numbers and whatnot.

vjr:
Updated the PR back to the original proposal like here: https://github.com/elementary/switchboard-plug-about/pull/252#issue-1395883580

about_si_unit

lenemter (dismissed):

vjr:
Thanks @lenemter for the approval. I'll wait for maybe @danirabbit to comment on whether this change gets merged?

src/Views/HardwareView.vala
```diff
@@ -583,6 +583,24 @@ public class About.HardwareView : Gtk.Grid {
...
         return hostname;
     }
+
+    // Format layperson-friendly size string, replacement for GLib.format_size ().
+    // Always return "GB", "TB" etc. even if IEC_UNITS requested, instead
+    // of "GiB", "TiB" etc. for the benefit of average users.
+    private string custom_format_size (uint64 size, bool iec_unit) {
+        uint divisor = iec_unit ? 1024 : 1000;
+
+        string[] units = { _("bytes"), _("KB"), _("MB"), _("GB"), _("TB"), _("PB")};
```
tintou, line 593: This should be `unowned string[]` and use ngettext to use the same strings as the ones available in GLib

src/Views/HardwareView.vala
```diff
@@ -583,6 +583,24 @@ public class About.HardwareView : Gtk.Grid {
...
+
+        int unit_index = 0;
+
+        while ((size / divisor) > 0 && (unit_index < units.length)) {
+            unit_index++;
+            size /= divisor;
+        }
+
+        return "%llu %s".printf (size, units[unit_index]);
```
tintou, line 602: I don't think that this will be great i18n wise

tintou (dismissed):
If the consensus is to merge it (which I'm still not convinced), here are a few things to fix before

vjr:
Thank you @tintou for the review.

I can look into address this over the weekend but might need help with the ngettext and i18n - anyone? Maybe @lenemter would you?

lenemter:
@vjr I've looked at GLib.format_size source code (https://github.com/GNOME/glib/blob/main/glib/gutils.c#L2837) and rewritten you function a bit for it to be as translatable as GLib one:

```
    private string custom_format_size (uint64 size, bool iec_unit) {
        uint divisor = iec_unit ? 1024 : 1000;

        string[] units = {_("KB"), _("MB"), _("GB"), _("TB"), _("PB")};

        int unit_index = 0;

        while ((size / divisor) > 0 && (unit_index < units.length)) {
            unit_index++;
            size /= divisor;
        }

        string unit;
        if (unit_index == 0) {
            // TRANSLATORS: Examples of use: "1 byte", "13 bytes"
            unit = ngettext ("byte", "bytes", (ulong) size);
        } else {
            unit = units[unit_index - 1];
        }

        /* TRANSLATORS: The first "%llu" is replaced with the value, the "%s" with a unit of the value.
           The order can be changed with "%$2s %$1llu". An example: "13 bytes" */
        return _("%llu %s").printf (size, unit);
    }
```

src/Views/HardwareView.vala
vjr: I get error `Invalid assignment from owned expression to unowned variable` if I use unowned keyword.

src/Views/HardwareView.vala
vjr: Applying code suggested by @lenemter ...

vjr:
Thanks @lenemter for the code! @tintou would you re-review?

lenemter (approved):

PR #253, opened 2022-10-30
Reboot to Firmware Setup

meisenzahl:
If I ever need to make changes on the firmware (BIOS/UEFI) of my device, I often don't know which key combination to press when starting the device to get there. Windows offers an option to boot directly into BIOS/UEFI. On Linux, I have used https://www.freedesktop.org/software/systemd/man/systemctl.html#--firmware-setup for this so far. This feature adds a button that reboots the device and starts the BIOS/UEFI.

Bildschirmfoto von 2022-10-30 05 55 54

Bildschirmfoto von 2022-10-30 05 56 11

### DIFF
--- a/src/Interfaces/LoginManager.vala
+++ b/src/Interfaces/LoginManager.vala
@@ -23,6 +23,7 @@
 public interface About.LoginInterface : Object {
     public abstract void reboot (bool interactive) throws GLib.Error;
     public abstract void power_off (bool interactive) throws GLib.Error;
+    public abstract void set_reboot_to_firmware_setup (bool interactive) throws GLib.Error;
 }
 
 public class About.LoginManager : Object {
@@ -67,5 +68,13 @@ public class About.LoginManager : Object {
         }
 
         return true;
+    }
+
+    public void set_reboot_to_firmware_setup () {
+        try {
+            interface.set_reboot_to_firmware_setup (true);
+        } catch (Error e) {
+            warning ("Could not connect to login interface: %s", e.message);
+        }
     }
 }
--- a/src/Interfaces/LoginManager.vala
+++ b/src/Interfaces/LoginManager.vala
@@ -71,15 +71,15 @@ public class About.LoginManager : Object {
         return true;
     }
 
-    public bool set_reboot_to_firmware_setup () {
+    public Error? set_reboot_to_firmware_setup () {
         try {
             interface.set_reboot_to_firmware_setup (true);
         } catch (Error e) {
-            warning ("Could not connect to login interface: %s", e.message);
-            return false;
+            warning ("Could not set reboot to firmware setup: %s", e.message);
+            return e;
         }
 
-        return true;
+        return null;
     }
 
     public bool can_reboot_to_firmware_setup () {
--- a/src/Interfaces/LoginManager.vala
+++ b/src/Interfaces/LoginManager.vala
@@ -24,6 +24,7 @@ public interface About.LoginInterface : Object {
     public abstract void reboot (bool interactive) throws GLib.Error;
     public abstract void power_off (bool interactive) throws GLib.Error;
     public abstract void set_reboot_to_firmware_setup (bool interactive) throws GLib.Error;
+    public abstract string can_reboot_to_firmware_setup () throws GLib.Error;
 }
 
 public class About.LoginManager : Object {
@@ -76,5 +77,15 @@ public class About.LoginManager : Object {
         } catch (Error e) {
             warning ("Could not connect to login interface: %s", e.message);
         }
+    }
+
+    public bool can_reboot_to_firmware_setup () {
+        try {
+            return interface.can_reboot_to_firmware_setup () == "yes";
+        } catch (Error e) {
+            warning ("Could not connect to login interface: %s", e.message);
+        }
+
+        return false;
     }
 }
--- a/src/Interfaces/LoginManager.vala
+++ b/src/Interfaces/LoginManager.vala
@@ -71,12 +71,15 @@ public class About.LoginManager : Object {
         return true;
     }
 
-    public void set_reboot_to_firmware_setup () {
+    public bool set_reboot_to_firmware_setup () {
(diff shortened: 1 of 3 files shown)


src/Views/OperatingSystemView.vala
```diff
@@ -125,17 +125,24 @@ public class About.OperatingSystemView : Gtk.Grid {
 
         var settings_restore_button = new Gtk.Button.with_label (_("Restore Default Settings"));
 
+        var reboot_to_firmware_setup_button = new Gtk.Button.with_label (_("Reboot to Firmware Setup"));
```
danirabbit, line 128: It looks like in the session indicator we're using "Restart" instead of "Reboot"
```suggestion
        var reboot_to_firmware_setup_button = new Gtk.Button.with_label (_("Restart to Firmware Setup"));
```

src/Views/OperatingSystemView.vala
```diff
@@ -125,17 +125,24 @@ public class About.OperatingSystemView : Gtk.Grid {
...
+
         var button_grid = new Gtk.ButtonBox (Gtk.Orientation.HORIZONTAL) {
             hexpand = true,
             layout_style = Gtk.ButtonBoxStyle.END,
             spacing = 6
         };
         button_grid.add (settings_restore_button);
+        button_grid.set_child_secondary (settings_restore_button, true);
+        if (LoginManager.get_instance ().can_reboot_to_firmware_setup ()) {
+            button_grid.add (reboot_to_firmware_setup_button);
```
danirabbit, line 138: Can we add this on the Firmware page instead of the Operating System page? It seems like that's where it's most relevant

src/Views/OperatingSystemView.vala
```diff
@@ -248,6 +257,33 @@ public class About.OperatingSystemView : Gtk.Grid {
         }
     }
 
+    private bool confirm_reboot_to_firmware_setup_action () {
+        var dialog = new Granite.MessageDialog.with_image_from_icon_name (
+            _("Are you sure you want to Reboot?"),
+            _("This will close all open applications and turn off this device."),
```
danirabbit, line 263: We can probably be more specific here
```suggestion
            _("Restart to firmware setup?"),
            _("This will close all open applications, restart this device, and open the firmware setup screen."),
```

src/Views/OperatingSystemView.vala
```diff
@@ -248,6 +257,33 @@ public class About.OperatingSystemView : Gtk.Grid {
...
+            "system-shutdown",
```
danirabbit, line 264: There's a more specific fd.o icon name for this:
```suggestion
            "system-reboot",
```

src/Views/OperatingSystemView.vala
```diff
@@ -248,6 +257,33 @@ public class About.OperatingSystemView : Gtk.Grid {
...
+            Gtk.ButtonsType.CANCEL
+        );
+        dialog.transient_for = (Gtk.Window) get_toplevel ();
+
+        var continue_button = dialog.add_button (_("Reboot"), Gtk.ResponseType.ACCEPT);
```
danirabbit, line 269: Let's use Restart to be consistent
```suggestion
        var continue_button = dialog.add_button (_("Restart"), Gtk.ResponseType.ACCEPT);
```

src/Views/OperatingSystemView.vala
```diff
@@ -248,6 +257,33 @@ public class About.OperatingSystemView : Gtk.Grid {
...
+        continue_button.get_style_context ().add_class (Gtk.STYLE_CLASS_DESTRUCTIVE_ACTION);
+
+        var result = dialog.run ();
+        dialog.destroy ();
+
+        return result == Gtk.ResponseType.ACCEPT;
```
danirabbit, line 275: `dialog.run ()` no longer exists in Gtk 4, so we shouldn't use it in new code. Instead we should call `dialog.present ()` and then connect to `dialog.response ()`

src/Views/OperatingSystemView.vala
```diff
@@ -248,6 +257,33 @@ public class About.OperatingSystemView : Gtk.Grid {
...
+    }
+
+    private void reboot_to_firmware_setup_clicked () {
+        if (confirm_reboot_to_firmware_setup_action ()) {
```
danirabbit, line 279: Kind of going with the other thing, we should connect to `dialog.response ()` since we can't get a return value from `run ()` anymore. This is kind of weird to have as a single-use function anyways

src/Interfaces/LoginManager.vala
```diff
@@ -68,4 +70,22 @@ public class About.LoginManager : Object {
 
         return true;
     }
+
+    public void set_reboot_to_firmware_setup () {
+        try {
+            interface.set_reboot_to_firmware_setup (true);
+        } catch (Error e) {
+            warning ("Could not connect to login interface: %s", e.message);
```
danirabbit, line 78: It seems like we don't want to fail silently here. If we can't set this flag then we should abort the system restart and throw an error to the user that we're unable to restart into the firmware setup screen. Otherwise we're restarting them for no reason with no explanation

danirabbit (changes requested):
Nice feature! I have a few inline comments

danirabbit:
I can confirm the feature works as expected, so happy to approve after the above is resolved :)

meisenzahl:
@danirabbit thank you very much for your feedback! I hope that I addressed all issues.

Bildschirmfoto von 2022-11-01 18 45 04

Bildschirmfoto von 2022-11-01 18 53 30

Bildschirmfoto von 2022-11-01 19 27 42

I was not sure if there is already an established way of handling dialogs as preparation for porting to Gtk4. Also feel free to improve the error dialog texts.

danirabbit:
@meisenzahl made a branch here :) https://github.com/elementary/switchboard-plug-about/pull/254

danirabbit (approved):
Let's go! Nice new feature :rocket:

PR #254, opened 2022-11-01
Simplify, show error message in dialog

danirabbit:
* Return the error message from LoginManager so we can display it in the dialog details
* Simplify single-use functions
* Use `dialog.present ()`
* Set dialogs modal
* Use more descriptive error dialog copy


meisenzahl (approved):
Thank you for your quick feedback. That makes it much easier :+1:

PR #258, opened 2023-01-26
OperatingSystemView: handle missing default wallpaper

danirabbit:
* Fixes a deprecation warning
* Use GLib to get system data dir
* Don't assume that a symbolic distributor logo exists
* Don't assume that the default wallpaper exists
* Fall back nicely


danirabbit:
@bobby285271 With this, would it make sense to drop the wallpaper build option or do you still want it?

bobby285271:
It looks like `Gtk.IconTheme.get_default ().has_icon (logo_icon_name + "-symbolic")` is `false` for us, so the build option should be a no-op for us right now (i.e. always no wallpaper). I cannot speak for the future and also cannot speak for other distros though.

<details>

Having elementary default wallpaper + NixOS default logo will be like:




Not sure how other distro think about this, but yeah I personally do think our logo doesn't play nice with the circle right now - our default logo is not a square one ([502×435px](https://github.com/NixOS/nixos-artwork/blob/master/logo/nix-snowflake.svg)) and also not a white one.

</details>

danirabbit:
@bobby285271 Yeah it shouldn't be possible to get that combination with this branch. The wallpaper background would only show if both a `-symbolic` version of the logo is available and a wallpaper matching `elementaryos-default` exists in system data dirs.

We can look at changing the name of that default in future branches if there was interest from downstreams in shipping a symbolic logo, but I'm thinking that this should do the thing you want without requiring a build flag. So if that makes sense to you, I'll remove the build flag in this branch as well

bobby285271:
:ok_hand: Dropping build flag should be fine to me

src/Views/OperatingSystemView.vala
```diff
@@ -39,39 +39,52 @@ public class About.OperatingSystemView : Gtk.Grid {
...
+                var file = File.new_for_path (
+                    Path.build_path (Path.DIR_SEPARATOR_S, path, "backgrounds", "elementaryos-default")
+                );
+
+                if (file.query_exists ()) {
+                    var file_icon = new FileIcon (file);
+
+                    var logo = new Hdy.Avatar (128, "", false) {
+                        // In case the wallpaper can't be loaded, we don't want an icon or text
+                        icon_name = "invalid-icon-name",
```
jeremypw, line 60: Do we need this line?  It just seems to produce lots of terminal critical warnings as "invalid-icon-name" cannot be found.  Leaving this line out had no noticeable effect for me.

jeremypw (changes requested):
One query inline but otherwise no regressions detected.

src/Views/OperatingSystemView.vala
danirabbit: Yeah that should be okay to remove since we're checking for the existence of the file before constructing this

jeremypw (approved):

PR #259, opened 2023-01-26
Use /usr/lib/upstream-os-release instead of /etc/upstream-release/lsb-release

tintou:
This is the place we want to use.


danirabbit (approved):
This looks good with the new base files :+1:

PR #260, opened 2023-01-28
HardwareView: Reuse translations from GLib

tintou:
Allows to not have duplicates of the string already availables.


ryonakano:
Hmm, the GLib translations referencing here look like introduced in 2.73.0: https://github.com/GNOME/glib/commit/05a8ef6647ae20bda17e76a3e01f4f0c805d3849

The GLib version on Horus is 2.72.4, so this takes effect from the next major version of elementary OS maybe? This would not be an issue though.

tintou:
@ryonakano Opened https://github.com/elementary/switchboard-plug-about/pull/265 for this, I really don't want translators to lose time on this :)

PR #265, opened 2023-02-24
HardwareView: Support GLib < 2.73 translation

tintou:
It is required as we _only_ have 2.72 in Jammy

### DIFF
--- a/src/Views/HardwareView.vala
+++ b/src/Views/HardwareView.vala
@@ -590,7 +590,11 @@ public class About.HardwareView : Gtk.Grid {
     private string custom_format_size (uint64 size, bool iec_unit) {
         uint divisor = iec_unit ? 1024 : 1000;
 
+#if HAS_GLIB_2_73
         const string[] UNITS = {"kB", "MB", "GB", "TB", "PB", "EB"};
+#else
+        const string[] UNITS = {"%.1f kB", "%.1f MB", "%.1f GB", "%.1f TB", "%.1f PB", "%.1f EB"};
+#endif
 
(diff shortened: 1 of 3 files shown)


ryonakano:
Tested with `LANGUAGE=ru io.elementary.switchboard` since it looks like the Russian uses different character with English alphabet for GB.

### Before (the latest master branch)
VirtualBox_elementary OS_24_02_2023_21_54_10

### After
VirtualBox_elementary OS_24_02_2023_21_53_42

ryonakano (approved):
Works as expected on Horus and code looks good, thank you for addressing this :tada:

src/Views/HardwareView.vala
```diff
@@ -590,7 +590,11 @@ public class About.HardwareView : Gtk.Grid {
     private string custom_format_size (uint64 size, bool iec_unit) {
         uint divisor = iec_unit ? 1024 : 1000;
 
+#if HAS_GLIB_2_73
         const string[] UNITS = {"kB", "MB", "GB", "TB", "PB", "EB"};
+#else
+        const string[] UNITS = {"%.1f kB", "%.1f MB", "%.1f GB", "%.1f TB", "%.1f PB", "%.1f EB"};
```
vjr, line 596: @tintou  @ryonakano  looks like this regresses the number format to show float instead of int? So it's showing `3.0 GB` instead of expected rounded `3 GB`

tintou: Unfortunately there is nothing we can do until we have a newer GLib and it's not a big deal even for casual users :wink:

PR #273, opened 2023-10-16
dialog.run () → dialogresponse.connect ()

danirabbit:
GTK 4 prep. No dialog.run () in GTK 4 so make sure we connect to response signal instead

### DIFF
--- a/src/Views/OperatingSystemView.vala
+++ b/src/Views/OperatingSystemView.vala
@@ -251,6 +251,7 @@ public class About.OperatingSystemView : Gtk.Grid {
         continue_button.get_style_context ().add_class (Gtk.STYLE_CLASS_DESTRUCTIVE_ACTION);
 
         dialog.response.connect ((response) => {
+            dialog.destroy ();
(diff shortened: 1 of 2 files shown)


src/Views/OperatingSystemView.vala
```diff
@@ -253,20 +250,18 @@ public class About.OperatingSystemView : Gtk.Grid {
...
 
-            foreach (var schema in all_schemas) {
-                reset_recursively (schema);
+                foreach (var schema in all_schemas) {
+                    reset_recursively (schema);
+                }
             }
-        }
+
+            dialog.destroy ();
```
jeremypw, line 262: This is slightly different behaviour from master ?  In master the dialog is destroyed before the schemas are reset in this branch the dialog is destroyed after.  Depending on how long the reset takes the dialog could appear not to have responded (not tested as I do not want to reset my settings!)

danirabbit: Fixed!

jeremypw (approved):

PR #280, opened 2024-01-22
Add system updates

leolost2605:
Some strings might need some refinement

Rquires elementary/settings-daemon#88


danirabbit:
Hey so I was playing around with a different design here: https://github.com/elementary/switchboard-plug-about/pull/281

danirabbit:
Yay, I'm glad we're back on one branch haha. I'll make another commit real quick to add back the cancel button!

danirabbit:
Okay, done pushing, I'll use branches if I want to propose something else, you can take back over :sweat_smile:

src/Views/OperatingSystemView.vala
```diff
@@ -95,40 +102,103 @@ public class About.OperatingSystemView : Gtk.Box {
...
+        updates_description = new Gtk.Label (null) {
+            xalign = 0
+        };
+        updates_description.add_css_class (Granite.STYLE_CLASS_SMALL_LABEL);
+        updates_description.add_css_class (Granite.STYLE_CLASS_DIM_LABEL);
+
+        var update_button = new Gtk.Button.with_label (_("Download")) {
+            margin_end = 6,
+            valign = CENTER
+        };
```
danirabbit, line 163: Maybe this should be suggested action after all

```suggestion
        };
        update_button.add_css_class (Granite.STYLE_CLASS_SUGGESTED_ACTION)
```

src/Views/OperatingSystemView.vala
```diff
@@ -95,40 +102,104 @@ public class About.OperatingSystemView : Gtk.Box {
...
+        update_button.add_css_class (Granite.STYLE_CLASS_SUGGESTED_ACTION)
```
danirabbit, line 164: We accidentally a semicolon

```suggestion
        update_button.add_css_class (Granite.STYLE_CLASS_SUGGESTED_ACTION);
```

src/Views/OperatingSystemView.vala
```diff
@@ -95,40 +103,116 @@ public class About.OperatingSystemView : Gtk.Box {
...
+            margin_top = 6,
+            margin_bottom = 6,
+            margin_start = 6
+        };
+        updates_grid.attach (updates_image, 0, 0, 1, 2);
+        updates_grid.attach (updates_title, 1, 0);
+        updates_grid.attach (updates_description, 1, 1);
+        updates_grid.attach (update_button_revealer, 2, 0, 1, 2);
+        updates_grid.attach (cancel_button_revealer, 3, 0, 1, 2);
+        updates_grid.attach (error_button_revealer, 4, 0, 1, 2);
```
danirabbit, line 205: Maybe if we're gonna have a button for every state we should put them in a stack :joy:

leolost2605: Would probably make sense :)

danirabbit:
@leolost2605 cleaned up some button stuff https://github.com/elementary/switchboard-plug-about/pull/283

leolost2605:
So I think I've got everything in that I wanted. Only missing from my side is an info button for package details and error info. Also I'm sure there are some refinements that can be done :)

danirabbit:
Yeah this is hella good! I'll put together a branch to show package details

src/Views/OperatingSystemView.vala
```diff
@@ -222,10 +319,123 @@ public class About.OperatingSystemView : Gtk.Box {
...
+        details_dialog.present ();
+    }
+
+    private async void refresh_clicked () {
+        if (update_proxy == null) {
+            return;
+        }
+
+        try {
+            yield update_proxy.check_for_updates (current_state.state == ERROR, false);
```
danirabbit, line 433: I'm not sure about this "force" behavior. This means that the refresh button doesn't have any effect sometimes. I think if someone requests a refresh we should always honor that

leolost2605: Hmm yeah I guess that makes sense. You can push if you want, I'm currently not at my computer :(

leolost2605: Done

leolost2605:
Let's go then! 🚀

PR #281, opened 2024-01-22
Updates Redesign

danirabbit:
Makes updates a feature of the operating system view instead of a whole separate view. Some things to maybe improve:
* Have a "show details" button that opens a dialog with a list of all the individual packages
* I'd rather use the update file size than number of packages but it's returning 0?
* Error handling

Screenshot from 2024-01-22 14 15 06


leolost2605:
First of all this looks really amazing!

> I'd rather use the update file size than number of packages but it's returning 0?

Yeah that's what I was going for but I don't know how to get it yet so that's why it's marked as todo in the settings daemon PR

> Error handling

So my current flow is that if an error occurs the settings daemon sends a notification that if clicked on spawns a message dialog whith details. Meanwhile the daemon also forces a cache refresh deleting all downloaded updates so that in the settings only the updates list gets shown again. So the settings wouldn't know about any error that happened. But maybe we want more something like
Error happened -> notification -> settings show error with info for details and refresh button

leolost2605 (approved):
Haven't looked at the code very much but since it's WIP anyways let's goo, it really looks amazing lol 😍

PR #292, opened 2024-01-25
OperatingSystemView: use urgent icon if updates contain security

### LINKED ISSUE #291: Use software-update-urgent if any security updates are included
It would be a nice touch to badge system updates with `software-update-urgent` instead of `software-update-available` when system updates includes security updates

danirabbit:
Fixes #291 

Depends on https://github.com/elementary/settings-daemon/pull/120

### DIFF
--- a/src/Views/OperatingSystemView.vala
+++ b/src/Views/OperatingSystemView.vala
@@ -370,6 +370,12 @@ public class About.OperatingSystemView : Gtk.Box {
                         details.packages.length
                     ).printf (details.packages.length);
 
+                    foreach (unowned var info in details.info) {
+                        if (info == SECURITY) {
+                            updates_image.icon_name = "software-update-urgent";
+                        }
+                    }
+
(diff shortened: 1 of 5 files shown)


src/Views/OperatingSystemView.vala
```diff
@@ -373,6 +373,12 @@ public class About.OperatingSystemView : Gtk.Box {
                         details.packages.length
                     ).printf (details.packages.length);
 
+                    foreach (unowned var info in details.info) {
```
leolost2605, line 376: I think you can simplify this to `if (Pk.Info.SECURITY in details.info)`

leolost2605 (changes requested):

src/Views/OperatingSystemView.vala
danirabbit: You're so smart!

leolost2605 (approved):
LGTM

PR #293, opened 2024-01-26
Clamp views instead of halign

danirabbit:
Screenshot from 2024-01-26 12 07 11


lenemter:
Firmware view doesn't use clamp. Is this intentional?

danirabbit:
@lenemter yeah it's a Granite.SimpleSettingsPage and I added clamps to Switchboard.SettingsPage so it'll get one automatically when it switches over

lenemter (approved):

PR #294, opened 2024-01-26
OperatingSystemView: Add a switch for automatic updates

danirabbit:
Goes with https://github.com/elementary/settings-daemon/pull/118

Screenshot from 2024-01-26 12 27 55

### DIFF
--- a/src/Views/OperatingSystemView.vala
+++ b/src/Views/OperatingSystemView.vala
@@ -183,6 +183,7 @@ public class About.OperatingSystemView : Gtk.Box {
         };
 
         var automatic_updates_header = new Granite.HeaderLabel (_("Automatic Updates")) {
+            hexpand = true,
             mnemonic_widget = automatic_updates_switch,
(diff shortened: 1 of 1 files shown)


src/Views/OperatingSystemView.vala
```diff
@@ -178,6 +178,24 @@ public class About.OperatingSystemView : Gtk.Box {
             child = details_button
         };
 
+        var automatic_updates_switch = new Gtk.Switch () {
+            valign = CENTER
+        };
+
+        var automatic_updates_header = new Granite.HeaderLabel (_("Automatic Updates")) {
```
leolost2605, line 185: I think `hexpand` would be good here to align the switch to the right like it's done with the action button above

```suggestion
        var automatic_updates_header = new Granite.HeaderLabel (_("Automatic Updates")) {
            hexpand = true,
```

leolost2605 (changes requested):

leolost2605 (approved):
LGTM :)

PR #299, opened 2024-02-10
Better (IMHO) icon :-)

vjr:
Use the `computer` icon instead of `application-x-firmware` which looks less busy to my tired old eyes.


vjr:
Screenshots:

Before:

about-before

After:

about-after

danirabbit:
I'm not sure this works because of the possible confusion with display settings. I think I'm :-1: here sorry

vjr:
No worries - do you have an alternative icon in mind? Or I can just close this PR.

danirabbit:
Not really I like the current icon :sweat_smile:

vjr:
alrighty then!

PR #300, opened 2024-02-17
Add driver view

leolost2605:
UI side of elementary/settings-daemon#132

Screenshot from 2024-02-18 18 52 10

I modeled this after the firmware view. Any suggestions are welcome
Unfortunately I think we can't do anything about the icons :(


src/Views/DriversView.vala
```diff
@@ -0,0 +1,170 @@
...
+    private Granite.Placeholder progress_placeholder;
+    private Granite.Placeholder error_placeholder;
+    private Drivers? driver_proxy;
+    private string error_message = "";
+
+    public DriversView () {
+        Object (
+            icon: new ThemedIcon ("application-x-firmware"),
+            title: _("Drivers"),
+            description: _("Additional drivers provided by device manufacturers can improve performance.")
```
danirabbit, line 20: I think we can copy the description text from Installer:
```suggestion
            description: _("Broadcom® Wi-Fi adapters, NVIDIA® graphics, and some virtual machines may not function properly without additional drivers.")
```

src/Views/DriversView.vala
```diff
@@ -0,0 +1,170 @@
...
+        );
+    }
+
+    construct {
+        var none_placeholder = new Granite.Placeholder (_("No drivers available")) {
+            description = _("Your system doesn't need any additional drivers."),
```
danirabbit, line 26: We usually try to avoid "you" forms for localization. I think also "system" is usually more used to refer to the operating system software, and "device" is usually more used to refer to the computer hardware

```suggestion
            description = _("This device doesn't need any additional drivers."),
```

src/Widgets/DriverRow.vala
```diff
@@ -0,0 +1,46 @@
...
+        var icon = new Gtk.Image.from_icon_name ("application-x-firmware") {
+            pixel_size = 32
+        };
+
+        var label = new Gtk.Label (driver_name) {
+            hexpand = true,
+            xalign = 0
+        };
+
+        var install_button = new Gtk.Button.with_label (installed ? _("Installed") : _("Install")) {
```
danirabbit, line 30: I think it might be more appropriate to use a radio here to make it more clear that this is a mutually exclusive selection

danirabbit (changes requested):
Very excited about this, fantastic work! Left a couple of inline comments.

For the overarching design direction, in looking at other driver managers it looks like maybe a better way to list drivers would be to have a toplevel item that represents the device and then a list of radio buttons that represent the available driver options. For example here's Ubuntu drivers:

software-updates-drivers-781529376

leolost2605:
@danirabbit ok I created a different branch (#309) with radio buttons. It (or rather settings daemon side of it) has a larger surface for errors so it should be tested carefully.

danirabbit:
Closing since we merged the other version

PR #301, opened 2024-02-18
OperatingSystemView: load logo asynchronoulsy when the plug was shown

### LINKED ISSUE #272: Wallpaper placeholder is visible behind elementary logo during navigation to plug
### What Happened?

When first opening this plug, the logo changes while the page is being animated in.

After the animation finishes, there is a faint inner "e" inside the outer "e". I don't think this is intentional?

https://github.com/elementary/switchboard-plug-about/assets/70772/578a24f5-2369-4a9f-bf5a-e282fb065c8c

### Steps to Reproduce

1. Open Switchboard
2. Open "About" plug

### Expected Behavior

The logo being shown is always the right one.

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

This is in a live session from the 2023-09-07 Daily ISO on a MacBook Pro.

leolost2605:
- Loads the wallpaper logo asynchronoulsy when the plug was shown
- Fixes a quite noticeable delay when clicking the plug until it's shown
- We begin loading only when the plug was actually shown because otherwise setting the paintable from data still makes the animation stutter
- Also fixes #272 by loading an empty paintable before replacing it

### DIFF
--- a/src/Plug.vala
+++ b/src/Plug.vala
@@ -23,6 +23,7 @@ public class About.Plug : Switchboard.Plug {
     private const string HARDWARE = "hardware";
     private const string FIRMWARE = "firmware";
 
+    private OperatingSystemView operating_system_view;
     private Gtk.Grid main_grid;
     private Gtk.Stack stack;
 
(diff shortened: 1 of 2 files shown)


danirabbit (approved):
Much faster, nice work!

PR #309, opened 2024-02-18
Add driver view v2 (with radio buttons)

leolost2605:
[As suggested](https://github.com/elementary/switchboard-plug-about/pull/300#pullrequestreview-1887302903) this groups drivers by device.
Since this [required](https://github.com/elementary/settings-daemon/pull/132#issuecomment-1950649413) quite a few more steps where different cli outputs have to be parsed, compared, compared again and then double checked, this PR (or rather the settings-daemon side of it) should be tested thoroughly. I've tested with a nvidia card and there it works however I don't know how it will behave with two or more devices etc.

Probably needs some UI refinements :)

Screenshot from 2024-02-18 23 33 10

elementary/settings-daemon#141


danirabbit (approved):
I can't test this myself but I think we should get it in for testing

PR #311, opened 2024-02-21
OperatingSystemView: use Gtk.Picture instead of Adw.Avatar

danirabbit:
Not sure if this is actually better. Gtk.Picure is expanding so I'm having to add a clamp which feels bad


jeremypw:
You can stop the distortion by setting a `column_spacing` on the `software_grid` of about `60` but this does not seem any better.

PR #315, opened 2024-03-25
OperatingSystemView: use boxed-list, modelbutton

danirabbit:
Screenshot from 2024-03-25 10 24 38

* Makes the whole auto updates row clickable
* Automatically styles separator, frame, etc


leolost2605 (approved):
Looks pretty slick :fire:

PR #316, opened 2024-03-31
Show a placeholder if no connection to the settings-daemon could be established

### LINKED ISSUE #312: Updates section empty if settings-daemon can't connect
### What Happened?

I somehow ended up in a state where settings daemon was unreachable and the updates area is just blank

### Steps to Reproduce

Probably kill settings daemon haha and then see:

Screenshot from 2024-03-14 10 59 57

### Expected Behavior

If we can't reach the settings daemon we should throw an error view and maybe prompt to log out to resolve it?

### OS Version

8.x (Early Access)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

leolost2605:
Fixes #312

### DIFF
--- a/src/Views/OperatingSystemView.vala
+++ b/src/Views/OperatingSystemView.vala
@@ -33,6 +33,7 @@ public class About.OperatingSystemView : Gtk.Box {
     private Gtk.Label updates_description;
     private Gtk.Revealer details_button_revealer;
     private Gtk.Stack button_stack;
+    private Gtk.Stack update_stack;
 
     construct {
         var style_provider = new Gtk.CssProvider ();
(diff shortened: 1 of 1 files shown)


src/Views/OperatingSystemView.vala
```diff
@@ -368,10 +380,13 @@ public class About.OperatingSystemView : Gtk.Box {
         try {
             current_state = yield update_proxy.get_current_state ();
         } catch (Error e) {
+            update_stack.visible_child_name = "placeholder";
```
danirabbit, line 383: Hm any reason not to use what's already there like:
```suggestion
                updates_image.icon_name = "dialog-error";
                updates_title.label = _("System updates not available");
                updates_description.label = _("Couldn't connect to the backend. Try logging out to resolve the issue.");
                button_stack.visible_child_name = "blank";
```

danirabbit (changes requested):

src/Views/OperatingSystemView.vala
leolost2605: Tbh I didn't even think of that :joy:

PR #321, opened 2024-06-02
HardwareView: Make hostname editable

### LINKED ISSUE #248: Feature to change Hostname via System->Hardware
### Problem

User can only change elementary hostname by editing the /etc/hostname through terminal.

### Proposal

Allow user to change hostname using GUI in Settings -> System -> Hardware

### Prior Art (Optional)

_No response_

ryonakano:
Fixes #248
Fixes elementary/switchboard-plug-sharing#53

Pressing Enter key in the entry now sets the hostname.

Currently we show the hostname only when the hardware is not made by OEM. So, I separate the widget that shows either hostname or the product name into two different widgets.

---

With OEM info, successfully acquired the Polkit permission to edit hostname:

スクリーンショット 2024-06-02 00 11 48

Without OEM info, failed to acquire the permission:

スクリーンショット 2024-06-02 00 13 06


src/Views/HardwareView.vala
```diff
@@ -554,6 +568,51 @@ public class About.HardwareView : Gtk.Box {
...
     }
 
+    private void request_hostname_permission () {
+        if (hostname_permission != null) {
+            return;
+        }
+
+        try {
+            // Asks for permission to execute SetStaticHostname and SetPrettyHostname
+            hostname_permission = new Polkit.Permission.sync (
```
danirabbit, line 578: We need to make sure we're setting both the static hostname and pretty hostname:

https://github.com/elementary/initial-setup/blob/c4108e028384c40ac31c5b0022e9224c1c314060/src/Views/AccountView.vala#L351

danirabbit (changes requested):
This looks good! Just one comment and I have a branch to add some more visual feedback here: #322

danirabbit:
I wonder if we also need to add something about a restart being required? It looks like for example Bluetooth sharing name isn't change until we restart

danirabbit (approved):
We can always revisit adding more feedback about restarting etc later. Let's get the functionality in for now :)

PR #322, opened 2024-06-03
Add feedback to host name entry

danirabbit:
* Add a save icon to the entry when its text is different from the current host name
* Add a spinner while we're setting the new hostname
* Add a checkmark icon when the new host name is saved

### DIFF
--- a/src/Views/HardwareView.vala
+++ b/src/Views/HardwareView.vala
@@ -157,7 +157,7 @@ public class About.HardwareView : Gtk.Box {
         hostname_entry.changed.connect (() => {
             if (hostname_entry.text != get_host_name ()) {
                 hostname_entry.secondary_icon_name = "document-save-symbolic";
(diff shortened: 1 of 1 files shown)


src/Views/HardwareView.vala
```diff
@@ -158,8 +154,31 @@ public class About.HardwareView : Gtk.Box {
             update_manufacturer_logo ();
         });
 
-        hostname_info.activate.connect (() => {
-            set_host_name.begin (hostname_info.text);
+        hostname_entry.changed.connect (() => {
+            if (hostname_entry.text != get_host_name ()) {
+                hostname_entry.secondary_icon_name = "document-save-symbolic";
+                hostname_entry.secondary_icon_tooltip_text = _("Update Device name");
```
ryonakano, line 160:
```suggestion
                hostname_entry.secondary_icon_tooltip_text = _("Update device name");
```

PR #325, opened 2024-06-10
OperatingSystemView: redesign links

### LINKED ISSUE #295: Revamp links to match Onboarding
### Problem

After onboarding goes away it's unclear how to get to these links again. I'm not sure there's another way to access the basics guide, for example

### Proposal

We should make the links here match the ones in Onboarding, perhaps using the same icons

### Prior Art (Optional)

Screenshot from 2024-01-26 12 32 32

danirabbit:
Fixes #295

Screenshot from 2024-06-10 14 25 10

* Use links from OSInfo where possible
* Add missing documentation link
* Replace translation link with Get Involved & Sponsor link
* Use Gtk.UriLauncher


src/Views/OperatingSystemView.vala
```diff
@@ -215,8 +247,55 @@ public class About.OperatingSystemView : Gtk.Box {
...
+        var thebasics_link = new LinkRow (
+            documentation_url,
+            _("Basics Guide"),
+            "text-x-generic-symbolic",
+            "green"
+        );
+
+        var support_link = new LinkRow (
+            support_url,
+            _("Get Community Support"),
```
danirabbit, line 259: I wonder if this should be "Get Help" instead? Is it just the word "Support" that implies its a support technician? Theoretically a distro could replace this link with a link to some paid support staff, so maybe it's not good to include "Community" here
```suggestion
            _("Get Help"),
```

src/Views/OperatingSystemView.vala
```diff
@@ -215,8 +247,55 @@ public class About.OperatingSystemView : Gtk.Box {
...
+            "help-contents-symbolic",
+            "blue"
+        );
+
+        var website_link = new LinkRow (
+            website_url,
+            _("Our Website"),
```
danirabbit, line 266: Is "Our Website" better than just "Website"? I'm not sure :sweat_smile:

zeebok (approved):

zeebok (approved):

PR #331, opened 2024-07-24
Plug: add more search terms

danirabbit:
Make sure this is updated for all the latest features/links/etc

### DIFF
--- a/src/Plug.vala
+++ b/src/Plug.vala
@@ -116,15 +116,24 @@ public class About.Plug : Switchboard.Plug {
             (Gee.EqualDataFunc<string>)str_equal
         );
 
-        search_results.set ("%s → %s".printf (display_name, _("Operating System Information")), OPERATING_SYSTEM);
-        search_results.set ("%s → %s".printf (display_name, _("Hardware Information")), HARDWARE);
+        search_results.set ("%s → %s".printf (display_name, _("Operating System")), OPERATING_SYSTEM);
+        search_results.set ("%s → %s → %s".printf (display_name, _("Operating System"), _("Automatic Updates")), OPERATING_SYSTEM);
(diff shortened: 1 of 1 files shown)


src/Plug.vala
```diff
@@ -116,15 +116,24 @@ public class About.Plug : Switchboard.Plug {
...
+        search_results.set ("%s → %s → %s".printf (display_name, _("Hardware"), _("Device name")), HARDWARE);
+        search_results.set ("%s → %s → %s".printf (display_name, _("Hardware"), _("Host name")), HARDWARE);
+
         search_results.set ("%s → %s".printf (display_name, _("Firmware")), FIRMWARE);
-        search_results.set ("%s → %s".printf (display_name, _("Restore Default Settings")), OPERATING_SYSTEM);
-        search_results.set ("%s → %s".printf (display_name, _("Suggest Translations")), OPERATING_SYSTEM);
-        search_results.set ("%s → %s".printf (display_name, _("Send Feedback")), OPERATING_SYSTEM);
-        search_results.set ("%s → %s".printf (display_name, _("Report a Problem")), OPERATING_SYSTEM);
-        search_results.set ("%s → %s".printf (display_name, _("Get Support")), OPERATING_SYSTEM);
-        search_results.set ("%s → %s".printf (display_name, _("Updates")), OPERATING_SYSTEM);
```
ryonakano, line 127: I think we may want to leave this term or change to `Install Updates` (or similar phrase), so that users can quickly reach the operating system view and check for updates or install them manually

danirabbit: There's still "Automatic Updates" so if you search for "Updates" that will show

ryonakano (approved):

PR #343, opened 2024-11-06
build: Fix build with fwupd 2.0.0

bobby285271:
- Generally based on [Arch's patch](https://gitlab.archlinux.org/archlinux/packaging/packages/switchboard-plug-about/-/commit/d3716abb1a37b88a427a4227bdfb63dcf915f757).
- The `fwupd_release_get_locations` usage is also inspired by [KDE discover](https://invent.kde.org/plasma/discover/-/blob/v6.2.3/libdiscover/backends/FwupdBackend/FwupdBackend.cpp?ref_type=tags#L203) and [GNOME software](https://github.com/GNOME/gnome-software/blob/011d52b3c3388ca88421edeae8045f24b4f60c9d/plugins/fwupd/gs-plugin-fwupd.c#L542).
- Offline support is [gone](https://github.com/fwupd/fwupd/commit/ee40786de8800201e1676355c0709e947bd95474).

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -13,7 +13,7 @@ datadir = join_paths(prefix, get_option('datadir'))
 libdir = join_paths(prefix, get_option('libdir'))
 
 glib_dep = dependency('glib-2.0', version: '>=2.64.0')
-
+fwupd_dep = dependency('fwupd')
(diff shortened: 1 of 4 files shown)


danirabbit (approved):
Looks reasonable to me!

PR #345, opened 2024-12-04
Fix Scroll Padding

alainm23:
Before
image

After
image


src/Views/OperatingSystemView.vala
```diff
@@ -321,16 +325,20 @@ public class About.OperatingSystemView : Gtk.Box {
             child = software_grid
         };
 
+        var main_box = new Gtk.Box (VERTICAL, 0) {
```
danirabbit, line 328: Do we need to add a box? Can we add margins to the clamp?

danirabbit (approved):
Good job!

PR #346, opened 2024-12-04
Plug: use Adw.ToolbarView

danirabbit:
To make sure the headerbar has a shadow when we scroll

Screenshot from 2024-12-04 12 44 27


alainm23 (approved):

alainm23:
Good job, I thought the same

PR #347, opened 2024-12-04
Show sponsor goal progress bar

### LINKED ISSUE #329: Show sponsor goal progress bar
### Problem

Want to drive engagement towards sponsorship

### Proposal

Show a progress bar representing our current sponsor goal completion. Can use a graphql query:

```
query {
  organization(login: "elementary") {
		sponsorsListing {
      activeGoal {
        percentComplete
      }
    }
  }
}
```

### Prior Art (Optional)

_No response_

alainm23:
Fixed #329

image


src/Views/OperatingSystemView.vala
```diff
@@ -713,4 +731,60 @@ public class About.OperatingSystemView : Gtk.Box {
...
+                transition_type = SLIDE_UP_DOWN
+            };
+            labels_stack.add_named (new Gtk.Label (_("Sponsor Us")) {
+                halign = START
+            }, "title");
+            labels_stack.add_named (new Gtk.Label (_("23% towards $7,500 per month goal")) {
+                halign = START
+            }, "goal");
+
+            var progress_bar = new Gtk.ProgressBar ();
```
danirabbit, line 766: Should probably use a LevelBar so you don't get the little glowy progress animation

alainm23:
I think I should also think about a case where the user does not have internet.

src/Views/OperatingSystemView.vala
```diff
@@ -713,4 +731,116 @@ public class About.OperatingSystemView : Gtk.Box {
...
+                                .get_object_member ("data")
+                                .get_object_member ("organization")
+                                .get_object_member ("sponsorsListing")
+                                .get_object_member ("activeGoal");
+
+                            int64 percent_complete = sponsors_listing.get_int_member ("percentComplete");
+                            double target_value = sponsors_listing.get_double_member ("targetValue");
+
+                            levelbar.value = percent_complete / 100.0;
+                            target_label.label = _("%s%% towards $%s per month goal")
```
danirabbit, line 834: There's actually a library we're supposed to use for formatting currency with localization: https://github.com/elementary/switchboard-plug-locale/blob/1c1b281bc56811d456338c7d64d967b0e60353ab/src/Widgets/Preview.vala#L64C13-L64C21

alainm23: I understand, but I don't know this library in detail and I don't know exactly how to use it.

src/Views/OperatingSystemView.vala
```diff
@@ -713,4 +731,113 @@ public class About.OperatingSystemView : Gtk.Box {
...
+                    var sponsors_listing = root.get_object ()
+                        .get_object_member ("data")
+                        .get_object_member ("organization")
+                        .get_object_member ("sponsorsListing")
+                        .get_object_member ("activeGoal");
+
+                    int64 percent_complete = sponsors_listing.get_int_member ("percentComplete");
+                    double target_value = sponsors_listing.get_double_member ("targetValue");
+
+                    levelbar.value = percent_complete / 100.0;
```
danirabbit, line 830: You know what might be kind of fun is setting this with an Adw.Animation

alainm23: What do you think of this?
[Kooha-2024-12-06-07-18-17.webm](https://github.com/user-attachments/assets/57cdae62-2243-4278-9311-a8d2c49464ea)

danirabbit:
@tintou any thoughts on how a github token should be stored?

src/Views/OperatingSystemView.vala
danirabbit: I did a branch to do it with Adw.Animation. It's a lot fewer lines and we can have a nice easing function https://github.com/elementary/switchboard-plug-about/pull/352

danirabbit (approved):
All the UI parts of this look good to me! Localization, dark mode, etc all seems good

I think I'd like another opinion about if there's a better way to store a token :sweat_smile:  But otherwise I'm happy with this :tada:

alainm23:
@danirabbit we have a problem with the type of currency, in your case that you live in the USA, the type of currency is the dollar, but for me that I live in Peru, it is the Sol, and it is giving me an incorrect value of the goal. 

image

danirabbit:
> I think the currency type should always be in dollars, because that is the currency github works with.


Ah shit. Yeah we want to localize delimiters and like the placement of the currency symbol, but not the currency type itself haha. I'll investigate more how to use this library

alainm23:
As far as I can see, the library only works taking into account the configured regional settings.

danirabbit:
> I think we should not get so complicated and just add the dollar sign before

Well there's that, and that's fine it can picked up by regular string translation. But we also need to localize the delimiters. In en_US there should be a `,` every 3 places. In some locales it's a `.` instead. So we need to at least format the number

davidmhewitt:
Yeah, I don't love the idea of having a token here at all, even if it only has public read only access.

The correct way to do this would probably be to have an endpoint on the website that fetches this data using a token we can keep secret and just returns the data publicly.

davidmhewitt:
Opened https://github.com/elementary/website/pull/3690 to add an endpoint returning this info to the website.

davidmhewitt:
The new API endpoint is now available on the website: https://elementary.io/api/sponsors_goal

src/Views/OperatingSystemView.vala
```diff
@@ -713,4 +731,125 @@ public class About.OperatingSystemView : Gtk.Box {
...
+            grid.attach (main_label, 1, 0);
+            grid.attach (details_revealer, 1, 1);
+            grid.attach (link_image, 2, 0, 2, 2);
+
+            child = grid;
+            add_css_class ("link");
+            get_goal_progress (level_bar, target_label);
+        }
+
+        private void get_goal_progress (Gtk.LevelBar levelbar, Gtk.Label target_label) {
```
tintou, line 794: I'd made this function async and take a GLib.Cancellable to cancel the send_and_read on plug exit.

danirabbit (approved):
It's very cute, it works as expected, I'm happy with this. Nice work!

PR #356, opened 2024-12-16
Add Update Progress Indicator

### LINKED ISSUE #310: Progress indicator when downloading updates
### Problem

When downloading quite a lot of updates the view redundantly shows "downloading" without any kind of indication of how long this might take.

Screenshot from 2024-02-20 19 45 20@2x

### Proposal

It would be nice to have a progressbar or at least like "10/200 MB" to have an idea that there is progress being made

### Prior Art (Optional)

_No response_

alainm23:
Fixed #310 
Requires: https://github.com/elementary/settings-daemon/pull/173

image


src/Views/OperatingSystemView.vala
```diff
@@ -567,6 +593,18 @@ public class About.OperatingSystemView : Gtk.Box {
...
+    public string get_progress_text () {
+        if (download_size_max == 0) {
+            return "";
+        }
+
+        uint64 downloaded_size = download_size_max - download_size_remaining;
+        double downloaded_mb = (double) downloaded_size / (1024 * 1024);
+        double total_mb = (double) download_size_max / (1024 * 1024);
+
+        return "%.2f MB / %.2f MB".printf (downloaded_mb, total_mb);
```
danirabbit, line 605: Should use https://valadoc.org/glib-2.0/GLib.format_size.html probably instead of assuming the file size will be in megabytes

src/Views/OperatingSystemView.vala
```diff
@@ -183,12 +188,24 @@ public class About.OperatingSystemView : Gtk.Box {
             xalign = 0
         };
 
+        update_progress_bar = new Gtk.LevelBar.for_interval (0, 100) {
```
danirabbit, line 191: This should be a progress bar since we are indicating progress rather than level :)

src/Views/OperatingSystemView.vala
```diff
@@ -547,9 +565,17 @@ public class About.OperatingSystemView : Gtk.Box {
...
+
+                download_size_remaining = current_state.download_size_remaining;
+                if (download_size_remaining > download_size_max) {
+                    download_size_max = download_size_remaining;
+                }
+
                 updates_image.icon_name = "browser-download";
                 updates_title.label = _("Downloading Updates");
-                updates_description.label = current_state.message;
+                updates_description.label = "%s %s".printf (current_state.message, get_progress_text ());
```
danirabbit, line 578: Probably should make sure to use tabular numbers so the label doesn't bounce around as values change:

`<span font-features='tnum'>%s</span>`

stsdc (approved):

alainm23:
I just did my latest tests and it works as expected.

image

stsdc:
For me the whole updates box changes width when the values are being updated. It happens when numbers change. The effect is the jumping width.

ryonakano (commented):
Thanks, looks awesome! I suggested some changes as a PR, would like you to take a look at #360.

ryonakano:
> For me the whole updates box changes width when the values are being updated. It happens when numbers change. The effect is the jumping width.

@stsdc Has the above comment been fixed in #359 already?

stsdc:
No, this PR was only for vertical changes.

For horizontal I proposed here in comments.

alainm23:
> No, this PR was only for vertical changes.
> 
> For horizontal I proposed here in comments.

Do you have a screenshot or any way to reproduce the bug?

stsdc:
I have posted screenshots in the review.

ryonakano:
@stsdc I couldn't find any screenshots that look like showing the above issue in this PR, I'm sorry but could you post the screenshots again?

stsdc:
Here is the original: https://github.com/elementary/switchboard-plug-about/pull/356#discussion_r1898659374

alainm23:
I just added the `wrap` property

stsdc (approved):

ryonakano (approved):
:+1: from me too. I'm also committing to this branch though :see_no_evil:

PR #358, opened 2024-12-27
SystemUpdate: Get last refresh time directly from Settings Daemon instead of via GSettings

### LINKED ISSUE #342: Refreshing update check does not refresh "Last refreshed"
### What Happened?

Manually checking whether there is a new update does not refresh the widget "memory" of when was the last check.

### Steps to Reproduce

Do not check for any update for at least an hour. (Untoggle automatic check too)
Go to System
"Last checked: XX mn ago"
Click refresh, wait for refresh
Unchanged: "Last checked: XX mn ago"

### Expected Behavior

"Last checked: Right now"

We could say this bug is not very refreshing

### OS Version

8.x (Early Access)

### Session Type

Secure Session (Wayland)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

ryonakano:
Fixes #342

See also: https://github.com/elementary/settings-daemon/pull/175#issue-2760622522

### DIFF
--- a/src/DBus/SystemUpdate.vala
+++ b/src/DBus/SystemUpdate.vala
@@ -27,5 +27,5 @@ public interface SystemUpdate : Object {
     public abstract async void cancel () throws DBusError, IOError;
     public abstract async void check_for_updates (bool force, bool notify) throws DBusError, IOError;
     public abstract async void update () throws DBusError, IOError;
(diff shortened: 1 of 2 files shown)


leonardo-lemos (approved):
LGTM

PR #362, opened 2025-02-08
Show update size in OperatingSystemView

vjr:
Goes hand-in-hand with https://github.com/elementary/settings-daemon/pull/179


danirabbit (approved):
Nice work!

PR #337, opened 2024-10-02
Add System Logs view

tintou:
Allows to graphically access the system logs.

This is a draft and I won't have time to polish it so if anyone want to give it some love!

PR #374, opened 2025-04-09
Add System Logs view

leolost2605:
Allows to graphically access the system logs. Based on #337 by @tintou 

We currently only show the logs from the current boot.
Allow searching and refreshing.
Load in chunks of 200 entries, load more when scrolled end is reached.

A chunk is split into "subchunks" that are defined by time instead of number of items. We allow 1 ms of sync loading before having to wait for the next idle. We do this until at least the number of items in a chunk (200) are loaded or eof is reached.

Performance should be pretty good but lmk if you run into issues.
Also some feedback for the UI would be appreciated :)
It seems columnview and scrolled window have some issues if there are too many headers visible at once but idk if we can do something about that (except for changing the UI ofc)

Screenshot from 2025-04-12 17 51 13@1x


stsdc:
This looks really good!
Actually, for a long time I thought of adding System/Boot logs to Monitor.

danirabbit:
Yeah actually I wonder if that might be more appropriate in Monitor? It might feel more natural if you're looking to debug/manage/monitor the system that logs could be a part of that process

stsdc:
I'm in favour of putting System Logs in Monitor. I think we will have more freedom there to fully develop this functionality.

leolost2605:
Makes sense to me :)

If someone wants to do it right now feel free. You should be able to use SystemdLogModel and SystemdLogEntry without modifications.
Otherwise I would take another look at it once monitor is ported to GTK4 since we do rely on a bunch of GTK4 widgets (mainly columnview) for good performance.

danirabbit:
Closing since there's https://github.com/elementary/monitor/issues/439 and I think we agreed system logs makes more sense there

PR #417, opened 2026-08-21
OperatingSystemView: set stack overflow visible

danirabbit:
Fixes a visual bug in OS 9 where button shadows are clipped


src/Views/OperatingSystemView.vala
```diff
@@ -215,6 +215,7 @@ public class About.OperatingSystemView : Gtk.Box {
 
         button_stack = new Gtk.Stack () {
             hhomogeneous = false,
+            overflow = visible,
```
ryonakano:
```suggestion
            overflow = VISIBLE,
```

Because this is an enum: https://valadoc.org/gtk4/Gtk.Overflow.VISIBLE.html

ryonakano (changes requested):

src/Views/OperatingSystemView.vala
danirabbit: Oops silly mistake. Sorry!

